Issue #52, opened 2020-11-16
Grafana throws ExpiredTokenException after a certain period


sounavo:
After setting up Sitewise plugin on Grafana it works for a while and then starts throwing "failed to fetch query data: ExpiredTokenException: status code: 403, request id: a9c25a27-c012-488d-a65a-18fc21a1939f" The error is gone after restarting the the ec2 running grafana. After a while the error comes back again.
![Screenshot 2020-11-16 at 19 13 07](https://user-images.githubusercontent.com/20790480/99291259-d1a16c80-283f-11eb-972a-25554e0915dd.png)
![Screenshot 2020-11-16 at 19 14 37](https://user-images.githubusercontent.com/20790480/99291361-fe558400-283f-11eb-81d4-757205310d14.png)

bymattoa:
likely need a refresh hook in the credentials logic, to attempt to reload the file/re-request the assume role

ryantxu:
Perhaps related to:
https://github.com/grafana/timestream-datasource/issues/11

And hopefully can be fixed with shared code.  Hopefully:
https://github.com/grafana/grafana-aws-sdk/tree/main/pkg/awsds

(This is copied from how core is doing cloudwatch)

ryantxu:
@sounavo can you try again?  this is most likely fixed 

See the instructions on: https://github.com/grafana/iot-sitewise-datasource#aws-iot-sitewise-datasource-development-guide for how to get the latest dev build

sounavo:
@ryantxu I still seem to be getting this issue after installing the new build. Happy to screenshare over slack if that helps
![image](https://user-images.githubusercontent.com/20790480/100086502-cd7fdb00-2e4d-11eb-93c7-ae30bda7f62f.png)
![image](https://user-images.githubusercontent.com/20790480/100086542-d96b9d00-2e4d-11eb-9380-ff11191394ff.png)

bymattoa:
Hey @sounavo, I'm working to repro this in my own environment.

Quick question - can you confirm your auth type setup? Are you using the [ec2 instance role](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/iam-roles-for-amazon-ec2.html) to assume the `Assume Role ARN`? See below:

![Screen Shot 2020-11-24 at 12 47 16 PM](https://user-images.githubusercontent.com/19557121/100150162-163a9280-2e54-11eb-9380-d90efdb2af81.png)

[...]

PR #56, opened 2020-11-17
auth: use shared library for settings and session cache

ryantxu:
The updated cloudwatch configs were moved to:
https://github.com/grafana/grafana-aws-sdk/tree/main/pkg/awsds

This is the first project actually trying to use this... so buyer beware!


bymattoa (approved):
lgtm!

as I said over chime - might be worth tinkering with the expiry might assist with solving issue #52 

https://docs.aws.amazon.com/sdk-for-go/api/aws/credentials/

ryantxu:
> might be worth tinkering with the expiry might assist with solving issue #52

I vote we give this a try and see if that is reproduced.  If it is, then we need to fix the issue in:
https://github.com/grafana/grafana-aws-sdk

PR #58, opened 2020-11-19
[WIP][Hierarchy explorer] adding hierarchy 'tree' view

bymattoa:
- Adding the hierarchy 'tree' view. Each hierarchy will contain a list of associated child assets, which contains control to select or focus on that asset.

The 'tree' view only exposes a root +/-1 level of assets. It shows the parents of the inspected asset + all hierarchies and children. 

Screen Shot 2020-11-19 at 3 53 15 PM

- Updating the model explorer view to use the `AssetList` component, instead of a Select drop down.

Screen Shot 2020-11-19 at 3 53 30 PM

TODO: Search.


ryantxu (approved):
lets commit and then have a few smaller cleanups

* only two tabs
* use same asset list for modal vs tree
* add search (i can look at this)

PR #61, opened 2020-11-22
parent as card

ryantxu:
Lists parents as cards rather than a dropdown.  Still a little funky, but good enough?

image

### DIFF
--- a/src/components/browser/BrowseHierarchy.tsx
+++ b/src/components/browser/BrowseHierarchy.tsx
@@ -73,21 +73,6 @@ export class BrowseHierarchy extends Component<Props, State> {
     this.setState({ search: event.currentTarget.value });
   };
 
-  renderParents() {
-    const { asset, parents } = this.state;
-    if (asset && parents?.length) {
-      return (
-        <Select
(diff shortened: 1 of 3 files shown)


src/components/browser/BrowseHierarchy.tsx
```diff
@@ -133,8 +118,16 @@ export class BrowseHierarchy extends Component<Props, State> {
       <>
         {asset ? (
           <>
+            {parents &&
```
bymattoa, line 121: We need some kind of label here for the parents list.

What would you think about differentiating between parent/hierarchy/child divs & cards by background color or an icon?

bymattoa (commented):
Pulled this in to take a look locally. Looks like 'browse' for parent doesn't load the correct child assets:

Screen Shot 2020-11-23 at 10 50 27 AM

bymattoa:
Screen Shot 2020-11-23 at 12 40 42 PM

bymattoa (approved):
lgtm. 

I think we should add some flavor to the assets cards to differentiate between parents/selected/children. Can look into in follow ups

PR #73, opened 2020-12-24
fixing styling on AssetHierarchyList containers

bymattoa:
When multiple hierarchies are present, the AssetHierarchyList elements have large margins between each. This makes it appear as the other hierarchies are not being loaded.

Changing height to auto with a 10px top margin.


ryantxu (approved):
sorry for the slow review!!!

PR #75, opened 2021-01-20
Edge: move edge config out of "common" config

ryantxu:
Modifies #76 so that the edge configuration is not in the "common" package.

* allows entering any region (not just from selection)
* shows a warning when the endpoint is not specified


src/common/regions.ts
```diff
@@ -0,0 +1,11 @@
+import { SelectableValue } from '@grafana/data';
+
+export const standardRegions: Array<SelectableValue<string>> = [
```
ryantxu, line 3: @bymattoa can you find the real list of valid regions?

package.json
```diff
@@ -12,10 +12,10 @@
   "author": "Grafana Labs <team@grafana.com> (https://grafana.com)",
   "license": "Apache-2.0",
   "devDependencies": {
-    "@grafana/data": "latest",
-    "@grafana/runtime": "latest",
+    "@grafana/data": "7.3.0",
```
ryantxu, line 15: 7.4 is "latest" now... but we should still depend on 7.3 until we actually want to force it

ryantxu:
@macrovve can you take a look and see if this looks reasonable?

macrovve:
For sure, it looks good to me.

PR #77, opened 2021-01-25
Variables: add variable support

ryantxu:
support for query variables. Related to #113 

image


ryantxu:
@bymattoa -- most of this PR is formatting changes due to an upgrade in prettier :(

src/variables.ts
```diff
@@ -0,0 +1,37 @@
...
+
+  query = (request: DataQueryRequest<SitewiseQuery>) => {
+    const { targets } = request;
+    if (targets && targets.length === 1) {
+      const query = targets[0];
+      if (query.queryType === QueryType.ListAssets || query.queryType === QueryType.ListAssetModels) {
+        return this.datasource.query(request).pipe(
+          mergeMap((rsp) => {
+            const assets = new DataFrameView<HierarchyInfo>(rsp.data[0]);
+            const data: MetricFindValue[] = assets.map((a) => ({
```
ryantxu, line 25: The default behavior just uses the string values as the variables... BUT for this, I assume we actually want the ids as the template value and the name as what is displayed

bymattoa: lgtm

bymattoa (commented):
We're also going to need to fetch property IDs (which are available on models) as well as asset IDs. How would a user go about mapping these?

src/variables.ts
bymattoa: We will also need to fetch property IDs. (See thread in conversation)

ryantxu:
This now uses the "Describe Asset" query to pick properties:

image

toddtreece:
i think #134 and #127 cover the changes needed for #113. closing this one

PR #78, opened 2021-02-01
[bug + chore] Region update fix + updating test data

bymattoa:
- Updating the PropertyQueryEditor to refresh the drop down assets when the user switches regions
- Updating test data with new Sitewise models
- Updating test data to be generated from the shared demo account

(sorry for the large PR!)


ryantxu (approved):
:see_no_evil:

src/components/query/PropertyQueryEditor.tsx
```diff
@@ -77,8 +77,9 @@ export class PropertyQueryEditor extends PureComponent<Props, State> {
     const { query } = this.props;
     const assetChanged = query?.assetId !== oldProps?.query?.assetId;
     const propChanged = query?.propertyId !== oldProps?.query?.propertyId;
-    if (assetChanged || propChanged) {
-      if (!query.assetId) {
+    const regionChanged = query?.region !== oldProps?.query?.region;
+    if (assetChanged || propChanged || regionChanged) {
```
ryantxu, line 81: Is this the one real change in the the PR?

bymattoa: Yes - just forcing an updating in that hook

PR #83, opened 2021-03-04
Update dependencies

ryantxu:
This syncs the go dependencies with current release (and is backwards compatible)

The latest toolkit has updated lint rules and defines the release pipeline


src/components/browser/BrowseHierarchy.tsx
```diff
@@ -90,9 +90,10 @@ export class BrowseHierarchy extends Component<Props, State> {
           <Input css="" value={search} onChange={this.onSearchChange} placeholder="search..." />
           <br />
 
-          {asset.hierarchy.map(h => {
+          {asset.hierarchy.map((h) => {
             return (
               <AssetHierarchyList
+                key={h.value}
```
ryantxu, line 96: this is the only real frontend change -- the lint rules complained that key was missing

All other frontend changes cosmetic (and automatic)

PR #87, opened 2021-03-30
use latest grafana/aws-sdk

sunker:
Please test and verify that your normal path for authenticating the plugin works. Please also verify that the query editor looks alright and it's possible to get data from the backend.

What has changed in the aws-sdk? 
* made grafana/ui and grafana/data dev dependencies
* remove dependency to grafana/runtime. instead reading config from grafanaBootData
* made it possible to opt out on ConnectionConfig legend header and endpoint field (a requirement for sigv4)
* add unit tests

The API has not changed, so there should be no need to make changes to the ConfigEditor.tsx. 

Please release the plugin when this pr has been merged to master, and then let Jon know the new version number.

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
   "author": "Grafana Labs <team@grafana.com> (https://grafana.com)",
   "license": "Apache-2.0",
   "devDependencies": {
-    "@grafana/aws-sdk": "0.0.293",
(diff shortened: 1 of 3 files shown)


ryantxu (approved):
LGTM

PR #93, opened 2021-04-27
Updating 'AUTO' resolution to pick resolution based on the time range and max data points

bymattoa:
This change updates the `AUTO` resolution for Aggregate queries to pick the best fit resolution, or switch to using the raw data (GetAssetPropertyValueHistory api) when appropriate. At a high level the `AUTO` resolution validates if the highest resolution will exceed the max data points, or take more than 4 requests to load the panel of data. If so, the resolution picker will move to the next lowest resolution and perform the same check. The picker terminates on `1d` resolution. 

- how should max pages be determined? `Answer:` set to 4 for now. Seeing 100-200ms per request. Keeps the average max time to load to <1s 
- how should interval be used to determine the correct resolution? `Answer:` max data points + time range is used to determine the best fit resolution for a query. 

https://user-images.githubusercontent.com/19557121/116596819-b8661700-a8d9-11eb-878d-3c22efb93790.mov


ryantxu (approved):
This looks great -- the only thing I might change is the field naming strategy.  In this PR, it has an aggregate field name "min, max, mean, etc" --- but when you zoom in it gets renamed to the property value.  Perhaps we should zoom into a field named "raw"?

bymattoa:
> This looks great -- the only thing I might change is the field naming strategy. In this PR, it has an aggregate field name "min, max, mean, etc" --- but when you zoom in it gets renamed to the property value. Perhaps we should zoom into a field named "raw"?

Updated

PR #95, opened 2021-05-12
Chore: use automatic instance manager

ryantxu:
update to the latest SDK and use the automatic instance management


pkg/server/server.go
```diff
@@ -15,26 +16,26 @@ import (
 )
 
 type Server struct {
-	Datasource Datasource
+	Datasource    Datasource
+	channelPrefix string
+	closeCh       chan struct{}
+	queryMux      *datasource.QueryTypeMux
```
ryantxu, line 22: Not used yet -- but these will be helpful when we support streaming

pkg/server/server.go
```diff
@@ -15,26 +16,26 @@ import (
 )
 
 type Server struct {
-	Datasource Datasource
+	Datasource    Datasource
+	channelPrefix string
+	closeCh       chan struct{}
```
ryantxu, line 21: Not used yet -- but these will be helpful when we support streaming

pkg/server/server.go
ryantxu, line 22: rather than build the map on every request, this saves builds it once on startup

ryantxu:
@bymattoa -- no rush on this.  It should not change any behavior, only shifts things to use a shared wrapper for instance management.  This at least removes one extra interface abstraction :man_shrugging:

bymattoa (approved):

PR #99, opened 2021-07-08
Allow query by property alias

### LINKED ISSUE #98: Support propertyAlias to fetch data
SiteWise GET APIs for fetching SiteWise data (i.e. https://docs.aws.amazon.com/iot-sitewise/latest/APIReference/API_GetAssetPropertyValueHistory.html) support propertyAlias as an alternative to using assetId + propertyId.

ryantxu:
Fixes #98

This adds `propertyAlias` to the UI and query model for anything that accept it

image

Still totally untested since I don't know any real alias values to try :)

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,8 +16,15 @@
     "@grafana/data": "7.5.0",
     "@grafana/runtime": "7.5.0",
     "@grafana/ui": "7.5.0",
-    "@grafana/toolkit": "latest",
+    "@grafana/toolkit": "^8",
+    "tslib": "2.2.0",
     "@types/lodash": "latest",
     "lodash": "latest"
+  },
+  "resolutions": {
+    "rxjs": "6.6.3"
+  },
+  "engines": {
+    "node": ">=14"
   }
 }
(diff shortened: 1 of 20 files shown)


atifali:
We have currently hardcoded the AssetProperty for this case that is used to form the frames in the backend. Will only work for floats as of now:

https://github.com/grafana/iot-sitewise-datasource/blob/8bbc4ee05e297b22ee8181247a0cfee87735ac2c/pkg/resource/sitewise.go#L31

atifali:
this is one ad-hoc way to infer the `dataType`:
https://github.com/grafana/iot-sitewise-datasource/blob/8751cee79e667207977bb622c779e39e9adcd75d/pkg/framer/property_value_history.go#L29

not ideal but works for now - will be nice to have the `DescribeAssetProperty` end-point working with `propertyAlias` only, and that way we could actually have all the other descriptions too - see: https://github.com/grafana/iot-sitewise-datasource/issues/98#issuecomment-892947756

ryantxu:
@atifali -- looks good!   Since I created the issue, I can not approve it :)

atifali (approved):
see: https://github.com/grafana/iot-sitewise-datasource/issues/98

Issue #100, opened 2021-07-20
Support LADP and Linux authentication when region is Edge.


macrovve:
SiteWise Edge now support LDAP and Linux authentication, SiteWise plugin should be able to support this as well. 

When customer want to access data in the edge, they need to call `/authenticate` API first to get temporary AWS credentials which will expire after 4h. And they will use OpsHub to download their certificate. With AWS credentials and certificate, they could call SiteWise data plane API to fetch data. 

There two changes need to be made on the plugin:
* Provide a way to allow customer to enter their LDAP/Linux credential
* Implement a mechanism to refresh credential every 4h
* Support AWS session token, It only supports AWS access key and secret key now.

atifali:
feature merged - closing this.

PR #101, opened 2021-07-21
feature: linux/ldap auth modes for Edge

atifali:
see #100


pkg/sitewise/auth.go
```diff
@@ -0,0 +1,114 @@
...
+	"time"
+
+	"github.com/grafana/iot-sitewise-datasource/pkg/models"
+)
+
+type EdgeAuthenticator struct {
+	Settings models.AWSSiteWiseDataSourceSetting
+}
+
+func (a *EdgeAuthenticator) Authorize(ctx context.Context) (models.AuthInfo, error) {
```
macrovve, line 21: nit: I assume `a` means `authenticator`, but it might seem cryptic out of context.

pkg/sitewise/auth.go
```diff
@@ -0,0 +1,114 @@
...
+	}
+
+	client := &http.Client{Transport: tr}
+
+	authEndpoint := a.Settings.AWSDatasourceSettings.Endpoint + "authenticate"
+	resp, err := client.Post(authEndpoint, "application/json", bytes.NewBuffer(reqBodyJson))
+	if err != nil {
+		return models.AuthInfo{}, err
+	}
+	fmt.Print(resp)
```
macrovve, line 94: Should we use log here

pkg/sitewise/datasource.go
```diff
@@ -23,13 +24,21 @@ type Datasource struct {
 }
 
 func NewDatasource(settings backend.DataSourceInstanceSettings) (*Datasource, error) {
-	cfg := client.AWSSiteWiseDataSourceSetting{}
+	cfg := models.AWSSiteWiseDataSourceSetting{}
 	err := cfg.Load(settings)
+	if err == nil {
+		err = cfg.Validate()
```
macrovve, line 30: Put this after 34, then we don't need check `err == nil`

pkg/sitewise/client/client.go
```diff
@@ -101,7 +102,7 @@ func GetClient(region string, settings AWSSiteWiseDataSourceSetting, provider aw
 		swcfg.Endpoint = aws.String(settings.Endpoint)
 	}
 
-	if settings.Region == "Edge" {
+	if settings.Region == models.EDGE_REGION {
```
macrovve, line 105: I didn't see the definition of `models.EDGE_REGION`. Is there any change not commited?

pkg/models/setting.go
```diff
@@ -1,15 +1,24 @@
...
 
 import (
 	"encoding/json"
 	"fmt"
+
 	"github.com/grafana/grafana-aws-sdk/pkg/awsds"
 	"github.com/grafana/grafana-plugin-sdk-go/backend"
 )
 
+const EDGE_REGION string = "Edge"
```
macrovve, line 11: Nit: I saw some place is using `models.EDGE_REGION`, should we keep consistent with this.

pkg/models/setting.go
```diff
@@ -27,13 +36,42 @@ func (s *AWSSiteWiseDataSourceSetting) Load(config backend.DataSourceInstanceSet
 		s.Profile = config.Database // legacy support (only for cloudwatch?)
 	}
 
+	// Make sure to set an auth mode
+	if s.Region == EDGE_REGION {
+		if s.EdgeAuthMode == "" {
+			s.EdgeAuthMode = EDGE_AUTH_MODE_DEFAULT
```
macrovve, line 42: `s.Region == EDGE_REGION && s.EdgeAuthMode == "" `

pkg/models/setting.go
```diff
@@ -27,13 +36,42 @@ func (s *AWSSiteWiseDataSourceSetting) Load(config backend.DataSourceInstanceSet
...
+
 	s.AccessKey = config.DecryptedSecureJSONData["accessKey"]
 	s.SecretKey = config.DecryptedSecureJSONData["secretKey"]
 	s.Cert = config.DecryptedSecureJSONData["cert"]
+	s.EdgeAuthPass = config.DecryptedSecureJSONData["edgeAuthPass"]
+	return nil
+}
+
+func (s *AWSSiteWiseDataSourceSetting) Validate() error {
+	if s.Region == EDGE_REGION {
```
macrovve, line 54: nit: check `s.region != EDGE_REGION` first, leave early rule :)

https://stackoverflow.com/questions/53115008/why-is-good-to-check-for-false-first-in-an-if-statement

macrovve (commented):
Overall is good to me

pkg/sitewise/auth.go
atifali: yep that would be better!

pkg/sitewise/client/client.go
atifali: it's within `pkg/models/setting.go` I believe

pkg/models/setting.go
atifali: thanks for nice tidbit!

pkg/sitewise/auth.go
```diff
@@ -0,0 +1,116 @@
...
+	authEndpoint := a.Settings.AWSDatasourceSettings.Endpoint + "authenticate"
+	resp, err := client.Post(authEndpoint, "application/json", bytes.NewBuffer(reqBodyJson))
+	if err != nil {
+		return models.AuthInfo{}, err
+	}
+
+	log.DefaultLogger.Debug("edge auth response: %s", resp)
+
+	res := make(map[string]string)
+	json.NewDecoder(resp.Body).Decode(&res)
```
ryantxu, line 99: why not just deserialize into `models.AuthInfo`?

no need to make a map, and then pull values out of it

pkg/sitewise/datasource.go
```diff
@@ -41,15 +43,35 @@ func NewDatasource(settings backend.DataSourceInstanceSettings) (*Datasource, er
...
+		updateAuth := func() error {
+			authInfo, err := edgeAuthenticator.Authorize()
+			if err == nil {
+				cfg.AccessKey = authInfo.AccessKeyId
+				cfg.SecretKey = authInfo.SecretAccessKey
+				cfg.SessionToken = authInfo.SessionToken
+				cfg.AuthType = awsds.AuthTypeKeys
+				waitTime = time.Until(authInfo.SessionExpiryTime)
+				log.DefaultLogger.Debug("should wait for: ", "time:", waitTime)
+				waitTime = 10 * time.Second
```
macrovve, line 57: Why 10 times here.

atifali: just trying to test something here for now - will clean this up next after some testing

pkg/sitewise/auth.go
atifali: TIL!

pkg/sitewise/datasource.go
```diff
@@ -23,13 +25,56 @@ type Datasource struct {
...
+		edgeAuthenticator := EdgeAuthenticator{
+			Settings: cfg,
+		}
+
+		var waitTime time.Duration
+
+		updateAuth := func() error {
+			authInfo, err := edgeAuthenticator.Authorize()
+			if err == nil {
+				cfg.AccessKey = authInfo.AccessKeyId
```
FZambia, line 51: Need to use mutex since reading and writing happens in differrent goroutines, very very rare of course, but :)

pkg/sitewise/datasource.go
```diff
@@ -23,13 +25,56 @@ type Datasource struct {
...
+			}
+			return err
+		}
+
+		err = updateAuth()
+		if err != nil {
+			return &Datasource{}, err
+		}
+
+		go func() {
```
FZambia, line 67: Need to cleanup this goroutine on Dispose

pkg/sitewise/auth_test.go
```diff
@@ -0,0 +1,28 @@
...
+
+func TestSimpleAuth(t *testing.T) {
+	t.Skip()
+	settings := models.AWSSiteWiseDataSourceSetting{
+		EdgeAuthMode: "linux",
+		EdgeAuthUser: "username",
+		EdgeAuthPass: "password",
+	}
+
+	settings.Endpoint = "https://localhost:80"
```
FZambia, line 19: Does this test supposed to work? Maybe use httptest here?

```
ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
		fmt.Fprintln(w, "Hello, client")
}))
defer ts.Close()

settings.Endpoint = ts.URL
```

pkg/sitewise/auth.go
```diff
@@ -0,0 +1,107 @@
...
+					}
+					opts.Intermediates.AddCert(cert)
+				}
+				_, err := certs[0].Verify(opts)
+				return err
+			},
+		},
+	}
+
+	client := &http.Client{Transport: tr}
```
FZambia, line 88: Need to add some timeout

pkg/sitewise/auth.go
```diff
@@ -0,0 +1,107 @@
...
+
+	"github.com/grafana/grafana-plugin-sdk-go/backend/log"
+	"github.com/grafana/iot-sitewise-datasource/pkg/models"
+)
+
+type EdgeAuthenticator struct {
+	Settings models.AWSSiteWiseDataSourceSetting
+}
+
+func (a *EdgeAuthenticator) Authorize() (models.AuthInfo, error) {
```
FZambia, line 22: I think better to call everything `Authenticator`, `Authenticate` or `Authorizer`, `Authorize`

pkg/sitewise/datasource.go
```diff
@@ -23,13 +25,56 @@ type Datasource struct {
...
 	}
-	sessions := awsds.NewSessionCache()
 
+	err = cfg.Validate()
+	if err != nil {
+		return nil, err
+	}
+
+	if cfg.Region == models.EDGE_REGION && cfg.EdgeAuthMode != models.EDGE_AUTH_MODE_DEFAULT {
+		// TODO: refresh session every 4h since creds expire
```
FZambia, line 41: I think comment is wrong as `waitTime = time.Until(authInfo.SessionExpiryTime)` is used?

pkg/sitewise/datasource.go
```diff
@@ -23,13 +25,56 @@ type Datasource struct {
...
+		if err != nil {
+			return &Datasource{}, err
+		}
+
+		go func() {
+			for {
+				log.DefaultLogger.Debug("wait time until next credential fetch: ", "time:", waitTime)
+				<-time.After(waitTime)
+				log.DefaultLogger.Debug("updating edge auth credentials now")
+				updateAuth()
```
FZambia, line 72: Maybe in case of error here we should try to update with some shorter interval?

FZambia: btw, one more option to consider - avoid goroutine at all, do `updateAuth()` on initialization, save `authInfo.SessionExpiryTime`, then inside `GetClient` check  SessionExpiryTime and try to `updateAuth` again if needed.

pkg/sitewise/datasource.go
FZambia:
```go
type Config struct {
	Data string
}

func NewClient(cfg Config) {
	fmt.Println(cfg.Data)
}

func TestConcurrentReadWrite(t *testing.T) {
	cfg := Config{
		Data: "test",
	}
	//var mu sync.Mutex
	done := make(chan struct{})
	go func() {
		//mu.Lock()
		cfg.Data = "1"
		//mu.Unlock()
		close(done)
	}()
	//mu.Lock()
	cfgCopy := cfg
	//mu.Unlock()
	NewClient(cfgCopy)
	<-done
}
```

If you run this (which is an close analogue I believe) with race detector:

```
go test -race
```

Then race detector will report data races. With mutex code uncommented race detector is happy.

atifali: thanks for the catch!

pkg/sitewise/auth_test.go
atifali: yeah with the right credentials and cert it should work as expected - only meant for dev work hence skipped - will add comments for that

pkg/sitewise/datasource.go
atifali: removing this

pkg/sitewise/datasource.go
atifali: yeah i think for now an else with a small waitTime would be okay - although the other approach seems simpler

pkg/sitewise/auth_test.go
ryantxu: using httptest is a great idea!  then we can actually test this function :)

pkg/sitewise/datasource.go
```diff
@@ -20,21 +23,86 @@ type invokerFunc func(ctx context.Context, sw client.SitewiseClient) (framer.Fra
...
+	clientGetter := func(region string) (swclient client.SitewiseClient, err error) {
+		swclient, err = client.GetClient(region, cfg, sessions.GetSession)
+		return
+	}
 
-	return &Datasource{
-		GetClient: func(region string) (swclient client.SitewiseClient, err error) {
-			swclient, err = client.GetClient(region, cfg, sessions.GetSession)
+	if cfg.Region == models.EDGE_REGION && cfg.EdgeAuthMode != models.EDGE_AUTH_MODE_DEFAULT {
+		edgeAuthenticator := dummyAuthenticator{ //EdgeAuthenticator{
```
ryantxu, line 51: test this manually with the dummyAuthenticator, and lets then commit with EdgeAuthenticator

In another PR, lets explore how to do this on demand rather than as a background task.

ryantxu (approved):

pkg/sitewise/auth_test.go
atifali: will do this in a separate PR

macrovve:
Hi, I just test this version and overall is great, expect one small bug. In the current version, you have to provide endpoint with the slash, like this `https://<ip>:443/`, otherwise it will fail.  
image

I think we could accept both ways `https://<ip>:443/` and `https://<ip>:443` or provide some instructions if the format is wrong.

macrovve:
I saw this [pr](https://github.com/grafana/grafana-aws-sdk/pull/23/commits/1e6d22bdc8ee12623959b1ae30802ec7fe322af1)  has added the support of session token, but it didn't change the UI to allow the customer enter their session token.

And SiteWise Edge doesn't provide permanent credential anymore, so I was thinking it may be better to remove `Standard` authentication mode until we implement the session token UI @atifali

atifali:
> Hi, I just test this version and overall is great, expect one small bug. In the current version, you have to provide endpoint with the slash, like this `https://<ip>:443/`, otherwise it will fail.
> image
> 
> I think we could accept both ways `https://<ip>:443/` and `https://<ip>:443` or provide some instructions if the format is wrong.

should be fixed now - urls with a trailing slash or not will work

atifali:
> I saw this [pr](https://github.com/grafana/grafana-aws-sdk/pull/23/commits/1e6d22bdc8ee12623959b1ae30802ec7fe322af1) has added the support of session token, but it didn't change the UI to allow the customer enter their session token.
> 
> And SiteWise Edge doesn't provide permanent credential anymore, so I was thinking it may be better to remove `Standard` authentication mode until we implement the session token UI @atifali

@macrovve how common is it that for permanent credentials the user is expected to provide the session token? (I am not super familiar with this, but from what I have seen only the key pair is enough for the queries to work in the case of regions other than `Edge`)

we had to make that change in the pr you mentioned for the linux/ldap modes to work since they explicitly required the session token associated with the keys that the `/authenticate` endpoint provides, for the subsequent queries to work. In all other cases we are assuming that is not required and hence an empty field would be fine. Is that not a correct assumption?

macrovve:
> > I saw this [pr](https://github.com/grafana/grafana-aws-sdk/pull/23/commits/1e6d22bdc8ee12623959b1ae30802ec7fe322af1) has added the support of session token, but it didn't change the UI to allow the customer enter their session token.
> > And SiteWise Edge doesn't provide permanent credential anymore, so I was thinking it may be better to remove `Standard` authentication mode until we implement the session token UI @atifali
> 
> @macrovve how common is it that for permanent credentials the user is expected to provide the session token? (I am not super familiar with this, but from what I have seen only the key pair is enough for the queries to work in the case of regions other than `Edge`)
> 
> we had to make that change in the pr you mentioned for the linux/ldap modes to work since they explicitly required the session token associated with the keys that the `/authenticate` endpoint provides, for the subsequent queries to work. In all other cases we are assuming that is not required and hence an empty field would be fine. Is that not a correct assumption?

Correct. If the session token is not provided, AWS will consider it as  long-term security credentials. Those credentials only have access key and secret key. In most time, it is associated with the user. User is used to log in the console and mostly like it has admin permission.  Obviously,  AWS won't recommend it,  so the common way we use is to create a role  with limited permission and use it to make programmatic requests for AWS resources. We could use this role to get temporary AWS credentials which has session token. 

Here is the doc about the [Using temporary credentials with AWS resources](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_temp_use-resources.html)

macrovve:
I just realized that if customer want to use Grafana plugin, they may not likely pick the temporary credentials, since they have to change it every time after it expires. 

I still think we need to provide this though

atifali:
lets merge this feature for now - and i will do a PR on the sdk for the `sessionToken` UI, since I do agree that it can be useful in certain cases.

macrovve:
Do you know when this version will be available on the Grafana public plugins' repo? Is there automatic workflow to that?

ryantxu:
@macrovve -- we can push a release later today

PR #109, opened 2021-11-25
fix: use theme2 and update deps

atifali:
this fixes: https://github.com/grafana/support-escalations/issues/1421


daniellee (approved):
LGTM

PR #123, opened 2022-03-31
Make asset descriptions optional

### LINKED ISSUE #122: Panel does not show the list of asset models when an asset model has no description
When an asset model in IoT SiteWise does not have any description, the panel with a **Query Type** on **List asset models** shows **No data in response**. When exploring this panel, the **Query error Metric request error** shows up.

These errors do not show when all the asset models have a description.

Minimal Reproducible Example:
- in the Console of IoT SiteWise, create the demo if there is no asset models, and a dummy model with an empty description,
- in a Grafana instance:
  - create the IoT SiteWise Data Source,
  - create a dashboard with a panel with **Query Type** as **List asset models** -> the panel says **No data in response**,
  - explore the panel -> the error **Query error Metric request error** shows.
- back in the Console of IoT SiteWise, add a description to the dummy asset model,
- refresh the page of the exploration of the panel -> the list of asset models shows, and the panel itself does not say **No data in response** anymore.

The description if an optional field of the asset models, the plugin should not show any error when this optional field is missing.

The Grafana logs contain these errors around the time when this issue occurs:
```
failed to query data: Failed to query data: rpc error: code = Canceled desc = context canceled

failed to query data: Failed to query data: rpc error: code = Unavailable desc = error reading from server: EOF
```

We think the plugin crashes on empty description.

ryantxu:
Fixes #122 

Currently when descriptions are missing, the query can crash -- this makes all descriptions optional


atifali:
so tested this out locally:

created model with no description
Screen Shot 2022-04-05 at 12 05 09 PM

on the current version of the plugin - can confirm we get the internal 500 as described in the issue
Screen Shot 2022-04-05 at 12 07 34 PM

with this change looks like we do see the list - which includes the model with no description as well
Screen Shot 2022-04-05 at 12 20 41 PM

atifali (approved):
looks good - merging now

Issue #113, opened 2022-03-01
Add Variable Query editor


tracfren:
There are no query types available for IoT SiteWise when defining a variable based on a query. Can we support querying by asset using the ListAssets API?

tracfren:
Desired UX: We want users to be able to pick an active asset (based on a query) or select from a list of currently running processes.

Preferred Solution: Define a variable in Grafana using a sitewise query. The result needs to be comma-separated name:value pairs with the name being the asset name and the value the asset ID.

taleena:
**Existing workaround:**
Query the data (batches, assets) from a different data source (e.g. Athena)  while populating a dashboard variable. Alternatively, hard code the asset IDs. This makes the deployment process more complex.

Of the recently reported issues, this is the 1st priority.

taleena:
Customer does not consider the workarounds as feasible, so we need to provide an estimated timeline.

lean257:
Hi @tracfren @taleena I just merged a change to this issue, would you be able to check and verify to close the issue? Thanks!

taleena:
I've sent a heads up to the requestor, haven't heard anything back yet but please feel free to proceed!

toddtreece:
this has been released in v1.3.0:
https://github.com/grafana/iot-sitewise-datasource/releases/tag/v1.3.0
https://grafana.com/grafana/plugins/grafana-iot-sitewise-datasource/

PR #127, opened 2022-05-02
Variables: Add variable query editor

lean257:
Related to #113 
Screen Shot 2022-05-02 at 6 34 28 PM


src/utils.ts
```diff
@@ -0,0 +1,23 @@
...
+export function frameToMetricFindValues(df: DataFrame): MetricFindValue[] {
+    const res:MetricFindValue[] = [];
+    const view = new DataFrameView<SimpleResult>(df);
+    view.forEach(item => {
+        const id = item.id;
+        const name = item.name;
+        if (id && name) {
+            res.push({text: name, value:id});
+        }
+        const value = id || name;
```
lean257, line 17: should this be inside an else so that we only push either or, instead of both scenarios?

ryantxu: it is weird -- in 7.5+, the DataFrameView would let us check fields outside the loop, but I don't think we want to bump the requirements for that.

I think this is OK as a first pass

src/components/query/QueryEditor.tsx
```diff
@@ -19,7 +19,12 @@ export const firstLabelWith = 15;
 
 export class QueryEditor extends PureComponent<Props> {
   onQueryTypeChange = (sel: SelectableValue<QueryType>) => {
-    const { onChange, query, onRunQuery } = this.props;
+    const { onChange, query } = this.props;
+    // hack to use QueryEditor as VariableQueryEditor
+    let { onRunQuery } = this.props;
+    if (!onRunQuery) {
+      onRunQuery = () => {};
+    }
```
ryantxu, line 27:
```suggestion
    const onRunQuery = this.props.onRunQuery ?? () => {};
```
maybe a little better 🤷

src/utils.test.ts
```diff
@@ -0,0 +1,145 @@
+import { dataFrameFromJSON, DataFrameJSON } from '@grafana/data';
+
+import { frameToMetricFindValues } from './utils';
+
+describe('Test utils', () => {
+  it('convert simple values', () => {
+    const df = dataFrameFromJSON(listAssetsResponse as DataFrameJSON);
+    expect(frameToMetricFindValues(df)).toMatchInlineSnapshot(`
```
ryantxu, line 8: hymm -- something funky here.  The result has 1 row, so I would expect 1 result 🤔

src/utils.ts
ryantxu: dooh -- yes obviously!!! 🤦 

good thing we are writing tests!!!

ryantxu (approved):
LGTM

PR #135, opened 2022-06-07
Add support for unified alerting queries

### LINKED ISSUE #132: Issues creating alerts
I'm trying to create alerts in Grafana based on SiteWise data. The dashboard panels all work a treat (with with the historical values and aggregates query types) but I am unable to use either of this to create alerts. When I attempt to make an alert I see this warning:

> Failed to evaluate queries and expressions: failed to execute conditions: input data must be a wide series but got type long (input refid)

Steps to reproduce:
1. Browse to `/alerting/new` 
2. Choose SiteWise as the query, choose either the property value history, or property aggregate option for type of query.
3. Choose a specific sitewise asset, then a property from the asset (type of property is "Double")
4. Warning message appears with message from above
5. Save the alert, which works - but the alert does nothing
6. Browse back to the alert, observe that it shows "no data" when trying to see the alert details

The environment is running on Grafana 8 on Ubuntu (based on the `latest-ubuntu` tag of the `grafana-oss` image). Any thoughts on how to resolve this are appreciated.

toddtreece:
This adds support for Unified Alerting by switching from long to wide time series frames. Fixes #132

Get property value aggregates:
image

Get property value history:
image

Get property value:
image

I can limit this change to alert queries, but the change also fixes time series rendering in panels and explore.

Before:
image

After:
image


pkg/server/handlers.go
```diff
@@ -63,6 +65,11 @@ func (s *Server) handlePropertyValueHistoryQuery(ctx context.Context, req *backe
 		return DataResponseErrorRequestFailed(err)
 	}
 
+	wide, err := data.LongToWide(frames[0], &data.FillMissing{Mode: data.FillModeNull, Value: math.NaN()})
```
ryantxu, line 68: this looks great -- but would love to see the test structure change to show we now return a wide frame

pkg/testdata/property-history-values-A.golden.jsonc
```diff
@@ -0,0 +1,142 @@
...
+//  
+//  Frame[0] {
+//      "type": "timeseries-wide",
+//      "custom": {
+//          "resolution": "RAW"
+//      }
+//  }
+//  Name: Demo Turbine Asset 1
+//  Dimensions: 2 Fields by 35 Rows
+//  +-------------------------------+--------------------------+
```
toddtreece, line 11: @ryantxu here's an example

lean257 (approved):
Pulled down and tested and worked for me!

Issue #121, opened 2022-03-10
Support for additional aggregation levels


taleena:
Enhancement request on behalf of @tracfren

Support for more aggregation levels beyond 1 minute, 1 hour,  and 1 day 

**Desired functionality:**
The step between those level is too steep (~2 order of magnitude). Having at least 10 sec, 10 minutes, 10 hours will allow for a smoother transition

**Suggested solution:**
Support all the levels that are supported by sitewise even if they are computed on-the-fly.

**Existing workaround:**
Settle for less visible information. Zoom in and pan to see more granular data

This is 1st on the list of nice to haves.

ryantxu:
Files to look at:
https://github.com/grafana/iot-sitewise-datasource/blob/main/pkg/sitewise/api/propvals/resolution_picker.go

https://docs.aws.amazon.com/iot-sitewise/latest/APIReference/API_GetAssetPropertyAggregates.html#API_GetAssetPropertyAggregates_RequestSyntax


When we wrote this, I think the resolutions were fixed -- if the API now supports more granular options then 👍

tracfren:
There is no API support for more granular annotations, but there is support for interpolated values: https://docs.aws.amazon.com/iot-sitewise/latest/APIReference/API_GetInterpolatedAssetPropertyValues.html.

toddtreece:
@tracfren are you suggesting that we use [GetInterpolatedAssetPropertyValues](https://docs.aws.amazon.com/iot-sitewise/latest/APIReference/API_GetInterpolatedAssetPropertyValues.html) for  10 sec, 10 min, & 10 hour intervals in the aggregate query type, or add a new interpolated query type?

tracfren:
@toddtreece I'm not 100% sure. I see pros/cons for both directions. We definitely want to call out that the values are interpolated and not confuse the customer about whether or not they are looking at real data. Also, the customer wants to be able to deep dive into the real/raw data often, as they are analytically focused. Do you have any recommendations of how we can make the switching between aggregate, interpolated, and raw data as seamless as possible?

toddtreece:
https://grafana.com/grafana/plugins/grafana-iot-sitewise-datasource/ `v1.4.0` supports interpolated queries with additional resolution levels. please let us know if that works for their use case

taleena:
Closing this as we haven't heard back, but the functionality has been added.

PR #136, opened 2022-08-09
Add support for interpolated property value queries

toddtreece:
### Why 

#121 requests additional aggregation levels, but additional levels are not supported by the SiteWise aggregation API.

### What

The current query editor would need to be rethought/redesigned to support dynamically switching between raw, aggregate, and linear interpolated queries without switching query types, so this matches the current pattern and adds interpolated property value queries as a new query type.

interpolated


ryantxu (approved):
🎉

PR #137, opened 2022-08-09
Add support for last observed value

### LINKED ISSUE #119: Replace missing value with the latest recorded value
Desired UX: When value is missing we want to be able to replace it with the latest known value and show it on the plot and use in alerts.

Preferred Solution: When defining a query, specify that a null should be replaced with the latest known value.

toddtreece:
## Why
Fixes #119 

>  The problem is, for example, you could be viewing the last hour of data, but the last point in SiteWise was ingested 1.5 hours ago and a line chart would show there is essentially no data. Depending on the sampling rate, this could be an incorrect assumption, as maybe the sampling rate is every 5 hours and the data point from 1.5 hours ago is still a valid measurement.

## What

Adds fallback when the original query returns an empty result.

locf


ryantxu:
This looks like it is working for "Get interpolated property values" -- but not regular property values.  If that is expected, we should just hide it in the UI.

The other weird part is that the we are showing timestamps when nothing is found... this causes odd zoom suggestions!

pkg/models/property.go
```diff
@@ -12,9 +12,10 @@ import (
 // This is done simply due to lack of solid generics support in golang.
 type AssetPropertyValueQuery struct {
 	BaseQuery
-	AggregateTypes []string `json:"aggregates,omitempty"` // Not used for the history query
-	Quality        string   `json:"quality,omitempty"`
-	Resolution     string   `json:"resolution,omitempty"`
+	AggregateTypes  []string `json:"aggregates,omitempty"` // Not used for the history query
+	Quality         string   `json:"quality,omitempty"`
+	Resolution      string   `json:"resolution,omitempty"`
+	LastObservation bool     `json:"lastObservation,omitempty"`
```
ryantxu, line 18:
```suggestion
	Interpolate string  `json:"interpolate,omitempty"`
```

Given that the property is defined as:
```LINEAR_INTERPOLATION | LOCF_INTERPOLATION```
https://docs.aws.amazon.com/iot-sitewise/latest/APIReference/API_GetInterpolatedAssetPropertyValues.html#API_GetInterpolatedAssetPropertyValues_RequestSyntax

maybe better to use a string?  in the UI and check we can still use the boolean toggle, but this will be nicer when they add smooth :)

toddtreece:
@ryantxu i think this is ready for another look

ryantxu:
Looks like the value is fetched by has an error:
image

ryantxu:
I pulled, and now fixed!

ryantxu (approved):

PR #140, opened 2022-09-12
Use AWS SDK to calculate user agent string

wbrowne:


ryantxu (approved):

wbrowne:
@ryantxu can we do the same for `twinmaker`? I just don't have the rights for pushing there

PR #141, opened 2022-09-23
Fix nil values in interpolated queries

toddtreece:


pkg/framer/property_interpolated.go
```diff
@@ -17,36 +17,36 @@ type InterpolatedAssetPropertyValue struct {
...
 
-	for i, v := range p.InterpolatedAssetPropertyValues {
+	for _, v := range p.InterpolatedAssetPropertyValues {
 		value := getPropertyVariantValue(v.Value)
 		if value == nil {
 			continue
 		}
-		timeField.Set(i, getTime(v.Timestamp))
-		valueField.Set(i, value)
+		timeField.Append(getTime(v.Timestamp))
```
toddtreece, line 48: @ryantxu i took another look at this because it was bothering me, and i'm not sure where i was logging last time  looked it this, but it seems to be on our end. the fields were initialized with a length and skipping them on line 46 when `value == nil` causes the bug we were seeing.

drew08t (approved):

Issue #157, opened 2023-01-27
Migrate to create-plugin


idastambuk:
We want to move to create-plugin.

Migration details here: https://github.com/grafana/create-plugin#migrate-your-existing-plugin

PR #159, opened 2023-02-02
Migrate to create-plugin

idastambuk:
Implements [#157](https://github.com/grafana/iot-sitewise-datasource/issues/157)


package.json
```diff
@@ -3,28 +3,79 @@
...
+    "typescript": "^4.4.0",
+    "webpack": "^5.69.1",
+    "webpack-cli": "^4.9.2",
+    "webpack-livereload-plugin": "^3.0.2"
   },
   "resolutions": {
     "rxjs": "6.6.3"
+  },
+  "dependencies": {
+    "@emotion/css": "^11.1.3",
```
iwysiu, line 74: There's this sneaky thing you have to do which is move all the dependencies listed in .config/webpack/webpack.config.ts back to devDependencies, since we should be using main grafana's versions of them

iwysiu (commented):
Ah, I see the typecheck addition has struck again (spellcheck doesn't like it)

package.json
idastambuk: Oof, didn't see this. Will correct it!

fridgepoet (approved):

Issue #171, opened 2023-03-10
When a variable is a string, unnecessary extra quotation marks are added


idastambuk:
**What happened**:
When a variable value is a string (and not array, for example), the template_srv runner will return the string wrapped in extra quotation marks:  
<img width="201" alt="Screen Shot 2023-03-09 at 6 27 56 PM" src="https://user-images.githubusercontent.com/16140639/224325192-acd49e0d-2dc0-4867-8d70-74c85d844451.png">

The cause of this is in core grafana -> formatter used from grafana/scenes. The fix to this has already [been merged in grafana/scenes](https://github.com/grafana/scenes/pull/46), this issue is just to track updates to it (it needs release of Scenes + updating core grafana with the new version). I will post updates here as I get them. 

**How to reproduce it (as minimally and precisely as possible)**:
1. Add a constant variable to a dashboard, for example, with the value of a single string "test"
2. Add another variable, in the editor select, select Get property value history, then select the variable from 1. 
3. Observe that the query request contains the interpolated variable, but wrapped in extra quotes

**Anything else we need to know?**:

**Environment**:
- Grafana version: 9.4
- Plugin version: 
- OS Grafana is installed on:
- User OS & Browser:
- Others:

idastambuk:
[The fix will be released in 9.5](https://github.com/grafana/grafana/pull/64609#pullrequestreview-1334871410), after that this package has to be updated with the latest grafana/ version

idastambuk:
Scratch my last sentence - this package doesn't need to be updated. 
9.5 is currently being rolled out with the new version of /scenes so I will close this

Issue #158, opened 2023-01-30
Cannot select defined hierarchies when using "List associated assets" query type with another variable


nyamad:
**What happened**:
In a dashboard, I want to create and use multiple variables which store asset lists corresponds to each asset hierarchy.
To achieve this, for example, 
1. Create variable `parent` using "List assets" as Query type and suitable asset model
2. Create variable `child` using "List associated assets" as Query type, `${parent}` as Asset, and suitable hierarchy
3. Create variable `grandchild` using "List associated assets" as Query type, `${child}` as Asset, and suitable hierarchy
4. ...

It worked as expected for plugin versions from 1.3.0 to 1.5.1. But when I tested this with 1.6.0, It didn't work.

**What you expected to happen**:
In the Show field, defined hierarchies can be selected even if another variable is specified in the Asset field.

**How to reproduce it (as minimally and precisely as possible)**:
1. Create an asset model like `Child Asset Model` on AWS IoT SiteWise.
2. Create another asset model like `Parent Asset Model` and define a hierarchy like `Parent-Child-Hierarchy` with the `Child Asset Model`.
3. Create some assets (like `Parent Asset A` and `Child Asset A`) for each asset model, then associate them using defined hierarchy.
4. On Grafana, go to any dashboard and open the Dashboard settings, then select Variables.
6. Create a variable like `parent` (with Label "Parent") using SiteWise data source and specify **List assets** in the Query type field and `Parent Asset Model` in the Model ID field, then click **Apply**.
7. Create another variable `child` and specify **List associated assets** in the Query type, `${Parent}` in the Asset field.
8. But on the Show field, I cannot select `Parent-Child-Hierarchy`, so I cannot get expected results for the variable.

**Screenshots**

![creating_variable_using_sitewise-datasource-plugin](https://user-images.githubusercontent.com/72601311/215363583-e1a0dd0a-fb97-4867-8500-4650a482693b.png)

**Anything else we need to know?**:

**Environment**:
- Grafana version: 9.3.4
- Plugin version: 1.6.0
- OS Grafana is installed on: Ubuntu 22.04
- User OS & Browser: Windows10 & Chrome
- Others:

Thanks in advance.

idastambuk:
I was able to reproduce this and it looks like a bug on our end when interpolating variables. @nyamad the only workaround I can think of is if you select the parent directly instead of the variable, is this manageable for now? In the meantime we're taking a look. 

Thanks for bringing this to our attention!

[...]

PR #172, opened 2023-03-22
Variables: Fix assetId field variable replacement

idastambuk:
When selecting variables for assetIds field, if a variable is a string, the JSON.parse() fails, so the variable is added with the value of its ID instead of being correctly interpolated. This fixes that and adds some tests for assetIds. 

Fixes [#158](https://github.com/grafana/iot-sitewise-datasource/issues/158)

This can be reproduced by following the steps from the original PR and with Sitewise from datasource provisioning. The assetId field in the query object in the network tab should contain all the values defined for the variables. 

Screenshot 2023-03-22 at 3 16 03 PM
Screenshot 2023-03-22 at 3 14 35 PM

This also fixes the child not appearing when  a parent is selected in the List associated assets.
Screenshot 2023-03-22 at 3 19 42 PM

Note: There's another [variable interpolation issue](https://github.com/grafana/iot-sitewise-datasource/issues/171) which comes from core grafana, the fix for which will be released in 9.5


src/DataSource.test.ts
```diff
@@ -0,0 +1,85 @@
...
+  access: 'direct',
+  ...overrides,
+});
+
+jest.mock('@grafana/runtime', () => ({
+  ...jest.requireActual('@grafana/runtime'),
+  getTemplateSrv: () => {
+    return {
+      getVariableName(variableId: string) {
+        return variableId.slice(1);
```
sarahzinger, line 25: I think I don't follow this slice, is it just generating a string? Would it work the same as if we returned the variableId?

sarahzinger (commented):
Cool stuff! Couple of ideas/questions, let me know whatcha think!

src/DataSource.test.ts
idastambuk: Youre totally right, this was initially gonna be a more complicated test so this can definitely be simplified now! Just pushed

kevinwcyu (approved):
LGTM

sarahzinger (approved):
🚀

PR #176, opened 2023-04-19
Update backend dependencies

fridgepoet:
Closes https://github.com/grafana/oss-plugin-partnerships/issues/175


fridgepoet:
The code coverage workflow seems to fail when the organization of the repo is not `pkg/<datasource>/<all other directories`

We'll have to solve that, but I propose doing that in another PR and bypassing this check for this PR.

kevinwcyu (approved):

PR #179, opened 2023-05-01
Add ability to perform property queries by only specifying a property alias

### LINKED ISSUE #169: Scope: Issue querying data by `propertyAlias` when the property is associated with an asset
**What happened**:
Cannot construct query based on `propertyAlias` for a property which has a `propertyAlias` but is also associated with a SiteWise asset.

Able to construct queries based on `propertyAlias` for any properties which contain a `propertyAlias` (which both modeled and unmodeled data supports)

kevinwcyu:

This PR adds the ability to create a query with just the Property Alias field for property value queries.

Fixes #169 

Here's a video demonstrating different property queries using a property alias:

https://user-images.githubusercontent.com/19530599/236067496-61a554ee-7f5e-4b3e-ae39-88f46f654c5b.mov

The datasource is configured to us-east-2 and is using our external dev account.

Here's a quick overview of the path of a `Property Value` query on the backend:
1. server handler receives query and unmarshals the query json https://github.com/grafana/iot-sitewise-datasource/blob/9ea952893e9da5f8f719ecc1dd462311d34ea3ac/pkg/server/handlers.go#L131-L156
2. datasource handler invokes the api handler to execute the query https://github.com/grafana/iot-sitewise-datasource/blob/9ea952893e9da5f8f719ecc1dd462311d34ea3ac/pkg/sitewise/datasource.go#L132-L136
3. the invoke function executes the api handler `invoker` and calls `frameResponse` to build up the dataframe https://github.com/grafana/iot-sitewise-datasource/blob/9ea952893e9da5f8f719ecc1dd462311d34ea3ac/pkg/sitewise/datasource.go#L88-L100
4. `frameResponse` calls the Frames method to generate the dataframe, which for a Property Value query is implemented here https://github.com/grafana/iot-sitewise-datasource/blob/9ea952893e9da5f8f719ecc1dd462311d34ea3ac/pkg/framer/property_value.go#L15
5. this Frame method gets some properties related to the query https://github.com/grafana/iot-sitewise-datasource/blob/9ea952893e9da5f8f719ecc1dd462311d34ea3ac/pkg/framer/property_value.go#L18 which ultimately uses the AssetIds from the original query to get information about the property (e.g. data type, unit of the data, etc.) https://github.com/grafana/iot-sitewise-datasource/blob/9ea952893e9da5f8f719ecc1dd462311d34ea3ac/pkg/resource/query_resouce_provider.go#L59.
6. then the data frames are generated https://github.com/grafana/iot-sitewise-datasource/blob/9ea952893e9da5f8f719ecc1dd462311d34ea3ac/pkg/framer/property_value.go#L23-L50

I mention all of this because one of the changes in this PR is to pass a pointer of the query into the handler as mentioned here https://github.com/grafana/iot-sitewise-datasource/pull/179/files#r1182767615 and in the handler the query gets the asset id and property id added to it https://github.com/grafana/iot-sitewise-datasource/pull/179/files#diff-254bac299433bd7cf6e08c61ee4d0d132c9a2262a5084d99a2ffb9a10e8cde84R49. This is done so that we can get the information about the property as mentioned in step 5 above. I think it should be fine to mutate the query, but wanted to call this out in case there is another way of doing it that doesn't require mutating the query.


pkg/sitewise/datasource.go
```diff
@@ -112,55 +112,55 @@ func (ds *Datasource) HealthCheck(ctx context.Context, req *backend.CheckHealthR
...
+	return ds.invoke(ctx, req, &query.BaseQuery, func(ctx context.Context, sw client.SitewiseClient) (framer.Framer, error) {
 		return api.GetInterpolatedAssetPropertyValues(ctx, sw, *query)
 	})
 }
 
 func (ds *Datasource) HandleGetAssetPropertyValueHistoryQuery(ctx context.Context, req *backend.QueryDataRequest, query *models.AssetPropertyValueQuery) (data.Frames, error) {
-	return ds.invoke(ctx, req, query.BaseQuery, func(ctx context.Context, sw client.SitewiseClient) (framer.Framer, error) {
-		return api.BatchGetAssetPropertyValues(ctx, sw, *query)
+	return ds.invoke(ctx, req, &query.BaseQuery, func(ctx context.Context, sw client.SitewiseClient) (framer.Framer, error) {
+		return api.BatchGetAssetPropertyValues(ctx, sw, query)
```
kevinwcyu, line 122: changed `*query` to `query`. this is to be able to add the AssetIds to the `query.BaseQuery` which is used later on when generating the frames:

https://github.com/grafana/iot-sitewise-datasource/blob/bee0f33038f7d829ff70f215ddcec4923f6dc7df/pkg/framer/property_value.go#LL18C38-L18C38

https://github.com/grafana/iot-sitewise-datasource/blob/bee0f33038f7d829ff70f215ddcec4923f6dc7df/pkg/resource/query_resouce_provider.go#L59

i wasn't sure if it's ok to do it this way by making the query mutatable.

fridgepoet: Hmm it seems like we shouldn't add any information to the query in those functions, but I cannot tell what could be a better solution.
Maybe adding something to the framer.Framer? (for example framer.AssetPropertyValue ?)

fridgepoet:
I tried to set up the same query as you with my account, but I'm getting a panic in the plugin right here https://github.com/grafana/iot-sitewise-datasource/blob/9ea952893e9da5f8f719ecc1dd462311d34ea3ac/pkg/framer/property_aggregate.go#L109

It says I have 3 resp.SuccessEntries, so it's trying to range over 3 elements, but then on that line there is only one element in `a.Request.Entries`

Any idea what could be causing that? I see you don't have the problem.

kevinwcyu:
> I tried to set up the same query as you with my account, but I'm getting a panic in the plugin right here
> 
> https://github.com/grafana/iot-sitewise-datasource/blob/9ea952893e9da5f8f719ecc1dd462311d34ea3ac/pkg/framer/property_aggregate.go#L109
> 
> It says I have 3 resp.SuccessEntries, so it's trying to range over 3 elements, but then on that line there is only one element in `a.Request.Entries`
> 
> Any idea what could be causing that? I see you don't have the problem.

it's to do with the resolution and time range of the query. the aws api call is paged and we just append the results to the `SuccessEntries` https://github.com/grafana/iot-sitewise-datasource/blob/9ea952893e9da5f8f719ecc1dd462311d34ea3ac/pkg/sitewise/client/client.go#L59

this causes a mismatch of expectations when building the frames because we assume the success entries is flat with each item representing a single asset https://github.com/grafana/iot-sitewise-datasource/blob/9ea952893e9da5f8f719ecc1dd462311d34ea3ac/pkg/framer/property_aggregate.go#L109-L110

this is an existing issue, i've created a ticket for it here https://github.com/grafana/iot-sitewise-datasource/issues/182

pkg/sitewise/datasource.go
fridgepoet: I have a new idea to do this --
it's also going to change the query within the Handle... functions but we will manipulate and return just copies of the queries. 
We can also abandon using the "invoke" methods in those specific handlers. This way we don't need to modify the signature of "invoke."

Going to write up the suggestion.

pkg/sitewise/api/property_aggregate.go
```diff
@@ -71,13 +91,18 @@ func aggregateQueryToInput(query models.AssetPropertyValueQuery) *iotsitewise.Ba
 	}
 }
 
-func GetAssetPropertyAggregates(ctx context.Context, client client.SitewiseClient, query models.AssetPropertyValueQuery) (*framer.AssetPropertyAggregates, error) {
+func GetAssetPropertyAggregates(ctx context.Context, client client.SitewiseClient, query *models.AssetPropertyValueQuery) (*framer.AssetPropertyAggregates, error) {
```
fridgepoet, line 94: I would suggest returning a modified query, like this:

```golang
func GetAssetPropertyAggregates(ctx context.Context, client client.SitewiseClient,
	query models.AssetPropertyValueQuery) (models.AssetPropertyValueQuery, *framer.AssetPropertyAggregates, error) {

	var (
		maxDps = int(query.MaxDataPoints)
	)

	modifiedQuery, err := getAssetIdAndPropertyId(query, client, ctx)
	if err != nil {
		return models.AssetPropertyValueQuery{}, nil, err
	}

	awsReq := aggregateQueryToInput(modifiedQuery)

	resp, err := client.BatchGetAssetPropertyAggregatesPageAggregation(ctx, awsReq, modifiedQuery.MaxPageAggregations, maxDps)

	if err != nil {
		return models.AssetPropertyValueQuery{}, nil, err
	}

	return modifiedQuery,
		&framer.AssetPropertyAggregates{
			Request: *awsReq,
			Response: iotsitewise.BatchGetAssetPropertyAggregatesOutput{
				SuccessEntries: resp.SuccessEntries,
				SkippedEntries: resp.SkippedEntries,
				ErrorEntries:   resp.ErrorEntries,
				NextToken:      resp.NextToken,
			},
		}, nil
}
```

See suggestion [here](https://github.com/grafana/iot-sitewise-datasource/pull/183/files#diff-af04a694ac71008929ba1d662736c9e3e3460d0f984bb22dbaa02b672c734f67R89-R90)

pkg/sitewise/api/property_history.go
```diff
@@ -41,21 +33,37 @@ func historyQueryToInput(query models.AssetPropertyValueQuery) *iotsitewise.Batc
 
 	entries := make([]*iotsitewise.BatchGetAssetPropertyValueHistoryEntry, 0)
 
-	for _, id := range query.AssetIds {
-		var assetId *string
-		if id != "" {
-			assetId = aws.String(id)
-		}
+	if query.PropertyAlias != "" {
```
fridgepoet, line 36: Would also suggest using a switch, see [here](https://github.com/grafana/iot-sitewise-datasource/pull/183/files#diff-68191c15fcbc403d2bd4bb32eb2caacf7ed821ad4d3f9f69f704e465a09d9672R36)

pkg/sitewise/api/property_history.go
```diff
@@ -65,12 +73,17 @@ func historyQueryToInput(query models.AssetPropertyValueQuery) *iotsitewise.Batc
...
 }
 
-func BatchGetAssetPropertyValues(ctx context.Context, client client.SitewiseClient, query models.AssetPropertyValueQuery) (*framer.AssetPropertyValueHistory, error) {
+func BatchGetAssetPropertyValues(ctx context.Context, client client.SitewiseClient, query *models.AssetPropertyValueQuery) (*framer.AssetPropertyValueHistory, error) {
 	var (
 		maxDps = int(query.MaxDataPoints)
 	)
 
-	awsReq := historyQueryToInput(query)
+	err := getAndSetAssetIdAndPropertyId(query, client, ctx)
```
fridgepoet, line 81: Similar to other handlers, would suggest returning a query and passing around the modified query. See [here](https://github.com/grafana/iot-sitewise-datasource/pull/183/files#diff-68191c15fcbc403d2bd4bb32eb2caacf7ed821ad4d3f9f69f704e465a09d9672R71-R72)

pkg/server/test/property_value_history_test.go
```diff
@@ -131,3 +134,120 @@ var getPropertyValueBoolean testServerScenarioFn = func(t *testing.T) *testScena
...
+			},
+		},
+		goldenFileName: "property-history-values-from-alias-timeseries",
+		handlerFn: func(srvr *server.Server) backend.QueryDataHandlerFunc {
+			return srvr.HandlePropertyValueHistory
+		},
+	}
+}
+
+var getPropertyValueBooleanFromAlias testServerScenarioFn = func(t *testing.T) *testScenario {
```
fridgepoet, line 217: I think we should move away from how the existing tests were written.
Here's my suggestion: https://github.com/grafana/iot-sitewise-datasource/pull/183/files#diff-3054cdb6a1d1aa7888cd1b888d5e0d9745012593389ed18e7728d97167806a93R222

kevinwcyu:
Change from this commit https://github.com/grafana/iot-sitewise-datasource/pull/179/commits/80f018af412dad169674e764e4166dfa4d2e8de2

previously the highlighted aggregate options only showed up if a propertyId was set in the query, but now it also shows with an alias (an alias identifies an assetId + propertyId).

aggregate options

pkg/sitewise/api/property_aggregate.go
```diff
@@ -71,13 +91,18 @@ func aggregateQueryToInput(query models.AssetPropertyValueQuery) *iotsitewise.Ba
...
 
 	var (
```
iwysiu, line 96: nit: I don't think this has to be a var block

pkg/sitewise/api/property_aggregate.go
iwysiu: +1 to this also. The function name doesn't imply that query will be changed and there's no real indication that its going to be, so as is it could cause some sneaky issues when debugging.

iwysiu (commented):
After taking a look, I second Shirley's takes on the switch cases and explicitly returning a modified query from the functions that need to to make it clearer to future maintainers. (Strictly speaking, public functions shouldn't change function signature for best Go coding practices, but this isn't really a consumed library so its fine). I would also like to +1 the fact that the sitewise tests are weird, and I didn't realize that they were like that.

pkg/testdata/prop-val-for-time-range-raw-data-for-time-range-A.golden.jsonc
```diff
@@ -16,15 +16,15 @@
...
-//  | 2021-02-01 17:30:00 +0100 CET | 28.287864119831497 | GOOD           |
-//  | 2021-02-01 17:35:00 +0100 CET | 28.345162289561085 | GOOD           |
-//  | 2021-02-01 17:40:00 +0100 CET | 28.57531512865568  | GOOD           |
-//  | 2021-02-01 17:45:00 +0100 CET | 28.779631112505346 | GOOD           |
-//  | 2021-02-01 17:50:00 +0100 CET | 28.746434495312684 | GOOD           |
-//  | 2021-02-01 17:55:00 +0100 CET | 28.815794249589597 | GOOD           |
-//  | 2021-02-01 18:00:00 +0100 CET | 29.67006840640403  | GOOD           |
-//  | 2021-02-01 18:05:00 +0100 CET | 29.504278603194113 | GOOD           |
-//  | 2021-02-01 18:10:00 +0100 CET | 29.32301703873501  | GOOD           |
+//  | 2021-02-01 08:30:00 -0800 PST | 28.287864119831497 | GOOD           |
```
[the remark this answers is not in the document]
kevinwcyu: > Is this just because the original "golden" files were generated in CET and you are coming from PST?

yep, it auto updates when i run the tests from vscode.

> I'm wondering if we really master these golden files. Are they supposed to exist so that in the future we are aware of any output changes? (understand this is probably out of scope of this PR)

yea, think they're supposed to make it easier to compare outputs so we don't have to handwrite the outputs to use for assertions.

kevinwcyu:
@fridgepoet @iwysiu thanks for the review. i've made changes to the tests and also changed the handlers to not modify the query and inlined the invoke function into the handlers. i think that's all the comments address. let me know if i missed anything else.

fridgepoet:
Gave you some feedback, but I don't want this to be blocking. I think the PR might benefit from being broken up -- feedback could be applied in subsequent PRs if you prefer.
And let me know if you would like me to do something.

fridgepoet (approved):

pkg/server/test/property_value_aggregate_test.go
```diff
@@ -1,64 +1,105 @@
...
+	require.Nil(t, err)
 
-	mockSw := &mocks.SitewiseClient{}
+	for i, dr := range qdr.Responses {
+		fname := fmt.Sprintf("%s-%s.golden", "property-aggregate-values", i)
+		experimental.CheckGoldenJSONResponse(t, "../../testdata", fname, &dr, true)
+	}
+}
 
+func Test_propertyValueAggregateFromAliasHappyCase(t *testing.T) {
```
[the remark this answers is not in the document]
kevinwcyu: thanks @fridgepoet, i missed that one switch case.

for the tests, let's pair on it in a separate pr.

Issue #170, opened 2023-03-07
Set Property Alias field to be interactive


iwysiu:
The `Property Alias` field in the property query editor should have the `interactive` property so that the link in the tooltip can actually be clicked. It's on line https://github.com/grafana/iot-sitewise-datasource/blob/fa1bd02b12b7d37bcb8d1c58b31ec4f93ef33cef/src/components/query/PropertyQueryEditor.tsx#L304

PR #180, opened 2023-05-02
Set property alias and query type to have interactive tooltips

katebrenner:
Closes https://github.com/grafana/iot-sitewise-datasource/issues/170 and does the same for `query type` InlineField


iwysiu (approved):
LGTM!

PR #184, opened 2023-05-09
Replace deprecated setVariableQueryEditor with CustomVariableSupport

### LINKED ISSUE #154: [Request]: Ability to filter assets on Asset Name
Filing this on behalf of a feature request.

Based on the information provided on [the plugin page](https://grafana.com/grafana/plugins/grafana-iot-sitewise-datasource/), it looks like the assets are being pulled and identified by their ID numbers, which are randomly generated UUIDs. Therefore, users are unable to filter using visible Asset Name, rather than Asset IDs.

**What are you really trying to do?**
Filter specific assets using filters or regex that gets applied to visible Asset Name.

**Context and constraints**
This is an important issue and blocker for the user.

**Describe what you've already tried and where it falls short**

1. Create IoT Sitewise assets
2. Create Sitewise plugin
3. Create dashboard
4. Create variable in dashboard using Sitewise data source
5. Select "List assets" query type (eg Asset 1, Asset 2, Asset 3) 
6. Input regex to attempt to filter the returned assets (eg only show Asset 2)
7. See that this does not return the expected results as it is filtering on ID, not name

Example, if I have 4 assets:

A_ABC1_XYZ1
A_ABC1_XYZ2
B_ABC2_XYZ1
C_ABC1_XYZ1

Tried to pull only the last two specifically using the full names, (B_ABC2_XYZ1|C_ABC1_XYZ1), but the result returned was "None".

**Additional context**
The assumptions of the regex functionality: It should help filter data containing some constant info/schema. UUIDs are generated totally randomly, so this functionality does not work.
The user wants to filter their data on a constant field (such as asset name) rather than a random string such as the ID.

**Environment**
- Grafana version: Amazon Managed Grafana v8.4.7
- Data source version: SiteWise v1.4.1
- OS Grafana is installed on:
- User OS & Browser:
- Product or feature area: 
- Impacted routes: 
- Others:

fridgepoet:
This replaces the deprecated `setVariableQueryEditor` with `CustomVariableSupport` modeled after what is used in the Athena datasource.

Fixes #154

I made an attempt to updated the deprecated setVariableQueryEditor. If I'm not mistaken in the build, I think it's working! I modeled it off of what is done in Athena.
Any suggestions are very welcome!

Here's the video of the change:
https://github.com/grafana/iot-sitewise-datasource/assets/4163034/40debff9-1617-499e-8622-3e56e79f142b

Now the regex filters by the name of the values for the regex.

### DIFF
--- a/src/DataSource.ts
+++ b/src/DataSource.ts
@@ -15,6 +15,7 @@ import { Observable } from 'rxjs';
 import { getRequestLooper, MultiRequestTracker } from 'requestLooper';
 import { appendMatchingFrames } from 'appendFrames';
 import { frameToMetricFindValues } from 'utils';
+import { SitewiseVariableSupport } from './variables';
 
 export class DataSource extends DataSourceWithBackend<SitewiseQuery, SitewiseOptions> {
   // Easy access for QueryEditor
@@ -24,6 +25,7 @@ export class DataSource extends DataSourceWithBackend<SitewiseQuery, SitewiseOpt
   constructor(instanceSettings: DataSourceInstanceSettings<SitewiseOptions>) {
     super(instanceSettings);
     this.options = instanceSettings.jsonData;
+    this.variables = new SitewiseVariableSupport(this);
   }
 
   /**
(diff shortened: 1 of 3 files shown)


iwysiu (approved):
lgtm!

kevinwcyu (approved):

Issue #188, opened 2023-05-15
Refactor tests for property queries and aggregate queries


kevinwcyu:
Work from this comment https://github.com/grafana/iot-sitewise-datasource/pull/179#discussion_r1190933637

Issue #189, opened 2023-05-16
Refactor tests


kevinwcyu:
Work from this comment https://github.com/grafana/iot-sitewise-datasource/pull/179#discussion_r1190933637

Sample PR: https://github.com/grafana/iot-sitewise-datasource/pull/190

Test files
- [ ] pkg/resource/caching_resource_provider_test.go
- [ ] pkg/server/test/describe_asset_model_test.go
- [ ] pkg/server/test/describe_asset_test.go
- [ ] pkg/server/test/list_asset_models_test.go
- [ ] pkg/server/test/list_assets_test.go
- [ ] pkg/server/test/list_associated_assets_test.go
- [ ] pkg/server/test/property_value_history_test.go
- [ ] pkg/server/test/property_values_for_time_range_test.go

Once test files are refactored, the following helpers, test data can be removed as they won't be used anymore
- [ ] pkg/sitewise/api/test_data_test.go
- [ ] pkg/testdata
```[tasklist]
### Tasks
```

fridgepoet:
If it's clear to you which tests remain to be refactored, would you add a checklist or something?

katebrenner:
closing as not planned

PR #190, opened 2023-05-16
Refactor tests for property queries and aggregate queries

kevinwcyu:
Refactors the test for property value and property value aggregate queries.

Fixes: #188

Changes are based on comments from https://github.com/grafana/iot-sitewise-datasource/pull/179#discussion_r1190933637.

Changes made:
- Update tests names
- Keep inputs/outputs closer to the test case
- Add assertions to make sure all the mocks are called
- Add assertions on the important arguments to the mocks

Notes:
- Tests still need to be updated for other test files. I've created issue here https://github.com/grafana/iot-sitewise-datasource/issues/189. Just a placeholder for now. It's probably easier to tackle all of these in smaller PRs.


fridgepoet:
Yowza! Nice job, looked like a lot of work. Pretty overwhelming to read still but I'm not sure what could make an end-to-end style test more readable. Another idea I have (future?) is to have smaller tests at each layer. The backend in this repo is a little bit "too much" but at least there is a method to the madness, so there are some distinct layers...

I hope, though, that your change will make some debugging easier one day.

Something I am missing is where the golden files are still getting modified. I see they were removed in these files you modified, but do you think there was some value to having the large sets of data? (I don't actually know if there were large sets in these tests.) And also were there any files we are no longer using now that we pulled some input into the test files? 

> Update Property method in pkg/resource/sitewise.go so it will call DescribeAssetPropertyWithContext as long as an asset id and property id is available. Previously, it would not call DescribeAssetPropertyWithContext if propertyAlias was not an empty string, but now that we get the asset id and property id from the alias, we can call DescribeAssetPropertyWithContext.

I think we should make this the focus of this PR if it's a change in the behavior, or at least add it as one of the main things the PR does. What do you think?

pkg/resource/sitewise.go
```diff
@@ -28,7 +28,7 @@ func (rp *SitewiseResources) Asset(ctx context.Context, assetId string) (*iotsit
 }
 
 func (rp *SitewiseResources) Property(ctx context.Context, assetId string, propertyId string, propertyAlias string) (*iotsitewise.DescribeAssetPropertyOutput, error) {
-	if propertyAlias != "" {
```
sarahzinger, line 31: Might be good to separate out changes to the codebase from refactoring tests wdyt?

pkg/server/test/property_value_aggregate_test.go
```diff
@@ -17,19 +20,40 @@ import (
...
+		mock.Anything,
+		mock.MatchedBy(func(input *iotsitewise.BatchGetAssetPropertyAggregatesInput) bool {
+			entries := *input.Entries[0]
+			return *entries.EntryId == "1assetid-aaaa-2222-bbbb-3333cccc4444" &&
+				*entries.AssetId == "1assetid-aaaa-2222-bbbb-3333cccc4444" &&
+				*entries.PropertyId == "11propid-aaaa-2222-bbbb-3333cccc4444" &&
+				*entries.AggregateTypes[0] == "SUM"
+		}),
+		mock.Anything,
+		mock.Anything,
```
sarahzinger, line 38: do you think it would be easier to read if we named some of these inputs? I'm not sure if this is stylistically weird to do with go but I guess it feels like it'd be easier to follow this test if we were like:
```
mockSw.On(
		"BatchGetAssetPropertyAggregatesPageAggregation",
		mockRequestContext,
		mockInput,
		mockPages,
		mockMaxResults,
	)
```

wdyt? Maybe I'm being silly/nitpicky?

pkg/server/test/property_value_test.go
```diff
@@ -44,36 +59,82 @@ func Test_getPropertyValueHappyCase(t *testing.T) {
...
+	expectedFrame := data.NewFrame("Demo Turbine Asset 1",
+		data.NewField("time", nil, []time.Time{time.Date(2021, 2, 1, 19, 20, 0, 0, time.UTC)}),
+		data.NewField("Wind Speed", nil, []float64{23.8}).SetConfig(&data.FieldConfig{Unit: "m/s"}),
+		data.NewField("quality", nil, []string{"GOOD"}),
+	)
+	if diff := cmp.Diff(expectedFrame, qdr.Responses["A"].Frames[0], data.FrameTestCompareOptions()...); diff != "" {
+		t.Errorf("Result mismatch (-want +got):\n%s", diff)
 	}
+
+	mockSw.AssertExpectations(t)
```
sarahzinger, line 85: sorry for the very basic questions on my part but like what does this assert?

sarahzinger (approved):
This looks cool to me but I'm also not very confident in my golang skills yet, so I defer to @fridgepoet on this one! Thanks for taking this on!

pkg/server/test/property_value_test.go
fridgepoet: It makes sure that all the all the declared methods ("mockSw.On("...")") were called.

Sometimes we declare those mocks but the test doesn't even use them which can be misleading. It's kind of an indirect way to assert some paths/coverage, I guess. 
I think some people might put these in everywhere, but I'm not sure it's always worth it. The reason I suggested it in these tests is that they're kind of "end to end," very long, complex, repetitive, and there are many to re-do. While pairing there were a few times we noticed that a test was written incorrectly and not testing what we thought.

pkg/server/test/property_value_aggregate_test.go
fridgepoet: These are just really huge tests since they're at the highest level (feels a bit end to end); I cannot yet find the ideal balance between something "digestable" and something where the reader doesn't get re-directed all the time to find the underlying values. 

It's my opinion that by not extracting the inputs into variables, we make it easier to connect the query input to what's being passed to some of the methods, and so on down the chain.

I think this test remains difficult to read, but mostly because is testing a very long chain of external API calls. Some ideas: maybe duplicate this test to only make assertions on one method at a time (so we can just use mock.Anything) or we could separate the tests and test at lower levels.

I _think_ this test a reasonable improvement over the previous input which was read from very far away files and sometimes quite lengthy, and an output which was also written to some file far away. I wouldn't be surprised to see further improvements iteratively as we dive into this repo more. 

Anyway, I will ultimately support whatever Kevin thinks is best!

pkg/resource/sitewise.go
kevinwcyu: Extracted changes here #192. Once that one is merged, I'll merge this PR. Temporarily switched the base branch from `main` for this PR to #192 so this only shows the refactoring of the tests.

pkg/server/test/property_value_aggregate_test.go
kevinwcyu: i think i'll leave it as is for now and we can iterate on further improvements later.

i think keeping it as `mock.Anything` has the advantage of being more clear since it's what's provided by the mocking library and also when I hover over it I see the following so i can see a particular argument isn't important for the given test case.

```
const mock.Anything untyped string = "mock.Anything"
Anything is used in Diff and Assert when the argument being tested shouldn't be taken into consideration.
```

also a more minor reason, if we ever modify the arguments to these functions, we'd need to check every single place to make sure name still makes sense.

iwysiu (approved):
Nice work! It is easier to read without having to cross reference golden files

PR #192, opened 2023-05-23
Fetch asset property info if assetId and propertyId are available

kevinwcyu:
Extracted from this PR https://github.com/grafana/iot-sitewise-datasource/pull/190/files#r1197884481

Update `Property` method in `pkg/resource/sitewise.go` so it will call `DescribeAssetPropertyWithContext` as long as an asset id and property id is available. Previously, it would not call `DescribeAssetPropertyWithContext` if `propertyAlias` was not an empty string, but now that we get the asset id and property id from the alias, we can call `DescribeAssetPropertyWithContext`. You can see the test output files now have the asset name and property names in them e.g. "Demo Turbine Asset 1" and "Average Wind Speed".


iwysiu (approved):
LGTM, assuming the answer to my question is yes

sarahzinger (approved):

Issue #182, opened 2023-05-04
Property aggregate query causes panic if the call to BatchGetAssetPropertyAggregatesPagesWithContext needs to be paged


kevinwcyu:
**What happened**:

Run a property aggregate query with resolution set to auto and increase the time range until it causes the call to [BatchGetAssetPropertyAggregatesPageAggregation](https://github.com/grafana/iot-sitewise-datasource/blob/9ea952893e9da5f8f719ecc1dd462311d34ea3ac/pkg/sitewise/client/client.go#L44) needs to be paged.

This causes a panic because there is a mismatch of expectations when building the frames because we assume the success entries is flat with each item representing a single asset https://github.com/grafana/iot-sitewise-datasource/blob/9ea952893e9da5f8f719ecc1dd462311d34ea3ac/pkg/framer/property_aggregate.go#L109-L110

See original comment on this issue here https://github.com/grafana/iot-sitewise-datasource/pull/179#issuecomment-1535064145

**What you expected to happen**:

**How to reproduce it (as minimally and precisely as possible)**:

**Screenshots**

**Anything else we need to know?**:

**Environment**:
- Grafana version:
- Plugin version:
- OS Grafana is installed on:
- User OS & Browser:
- Others:

sarahzinger:
Moving this to Next since I see a [user](https://github.com/grafana/iot-sitewise-datasource/issues/195) is running into this issue

kevinwcyu:
this should be fixed by #194 as well

PR #194, opened 2023-06-01
Fix handle expression queries with more than 250 data points

### LINKED ISSUE #160: Failed to fetch query data: Data not showing for query type 'get property value aggregates' and 'get property value history'
Different issues with grafana-iot-sitewise-datasource when using _get property value aggregates_ and _get property value history_ query types. Really like the ability to zoom out and quickly get an overview of the data for a longer timeframe, and zoom in to relevant time intervals. Sometimes not all the data is being visualized, and other times nothing is showing for unknown reasons, since data is available through other APIs, usch as boto3 or directly from AWS CLI. 

I understand it is a lot of information below, so feel free to reach out if something does not make sense or you have any other questions. 

Thank you!

**What happened**:
Different versions of _grafana-iot-sitewise-datasource_ and with query type _get property value aggregates_ and _get property value history_ does not show any data, or all the data when data is availble and can be obtained through either [AWS CLI](https://docs.aws.amazon.com/cli/latest/reference/iotsitewise/index.html) or [boto3](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/iotsitewise.html)

Expect the available data in the selected timeframe to be displayed in the panels. 

There are different issues, but below is a more detailed description of the different cases, including screenshots and descriptions on how to replicate. 

**Environment**:
- Grafana version: Different ones
- Plugin version: 1.4.0 - 1.6.0

# Longer description of the problem we are facing
A document to state issues with the different versions of IoT SiteWise datasource, with different test cases. 

## General comments
- Version 1.6.0 are not stable at all, and continuously show error messgae "an error occured within the plugin". While having the same timeframe, could get data, some data or no data just by pressing the refresh dashboard a few times. Seem to have greater difficulty when there is more than 250 data points in the query. 
- Seems like expression limits the number of data points, and usually stops at 250 (which suscipscously is the same number as the maxResults in the [API](https://docs.aws.amazon.com/iot-sitewise/latest/APIReference/API_GetAssetPropertyValueHistory.html#API_GetAssetPropertyValueHistory_RequestSyntax).
- Quickly checked version 1.4.0, which had similar if not same results as 1.4.1.
- Aggregates seem to break if the first query would result in an empty response but _with_ a nextToken for values with a more recent timestamp.
- The number of total data points and in the first response/batch from query is dependent on the data source and timeframe. The source for data used in the tests was created once a minute, and therefore the timeframe/duration and number of was often the same. 

## Test 1 with data points on both side of desired timeframe
For this test, ensure that there exists data points before and after time interval. Meaning it should be a subset like the figure shows below: 

image

#### Test 1.1.1.1 - single query
Aggregate resolution 'AUTO' in Grafana or '1m' in API, and duration of 250 minutes (250 points)
|Checks             |   Grafana |    Boto3  |   AWS CLI |
|--                 |-----------|-----------|-----------|
|Total # points     |250        |250        |250        |
|# points 1st batch | 250(?)    | 100       |250        | 
|nextToken          |NA         |NA         |NA         |
| success?          | Yes       |Yes        |Yes        |

Aggregate resolution 'AUTO' in Grafana or '1m' in API, and duration of 251 minutes (251 points)
|Checks             |   Grafana |    Boto3  |   AWS CLI |
|--                 |-----------|-----------|-----------|
|Total # points     |251        |251        |251        |
|# points 1st batch | 250(?)    | 100       |251        | 
|nextToken          |Yes        |Yes         |NA         |
| success?          | Yes       |Yes        |Yes        |

#### Test 1.1.1.2 - single query + expression
Add an expression (e.g. Math operation, $A+1)
Adding an expression limits the number of values in the response. 
    
Aggregate resolution 'AUTO' in Grafana or '1m' in API, and duration of 311 minutes (311 points)
|Checks             |   Grafana |    Boto3  |   AWS CLI |
|--                 |-----------|-----------|-----------|
|Total # points     |250        |311        |311        |
|# points 1st batch | 250(?)    | 100       |311        | 
|nextToken          |Yes        |Yes         |NA         |
| success?          | **No**       |Yes        |Yes        |

Description if not succesful: stops after 250 data points, and graph does not show any more.

#### Test 1.1.2.1 - single query
Usually works, but depending on the timeframe (Get different number of results by selecting '<' or '>' next to the timeframe). Sometimes it breaks even though property history value indicates there is data. 

#### Test 1.1.2.1 - single query + expression

Same as previous version. Stops after 250 data points, and graph does not show any more. 

image

Figure illustrates the problem. Panel on top is 'Stat' with 'Count' calculation. Panel on bottom is timeseries with a math expression ($A-1000)

#### Test 1.1.3.1 - single query
Same as version 1.5.0

#### Test 1.1.3.2 - single query + expression
Same as before, breaks at 250. 

### Test 1.2.1 - Grafana IoT SiteWise datasource version 1.4.1

Get Property Value History and duration of 250 and 251 minutes for our case (1 test point a minute maxResults from API is 250)

#### Test 1.2.1.1 - single query

Duration of 250 minutes (250 points)
|Checks             |   Grafana |    Boto3  |   AWS CLI |
|--                 |-----------|-----------|-----------|
|Total # points     |250        |250        |250        |
|# points 1st batch | 250(?)    | 100       |250        | 
|nextToken          |No         |Yes        |NA        |
| success?          | Yes       |Yes        |Yes        |

Duration of 251 minutes (251 points)
|Checks             |   Grafana |    Boto3  |   AWS CLI |
|--                 |-----------|-----------|-----------|
|Total # points     |251        |251        |251        |
|# points 1st batch | 250(?)    | 100       |251        | 
|nextToken          |Yes         |yes         |NA         |
| success?          | Yes       |Yes        |Yes        |

#### Test 1.2.1.1 - single query + expression

|Checks             |   Grafana |    Boto3  |   AWS CLI |
|--                 |-----------|-----------|-----------|
|Total # points     |250        |311        |311        |
|# points 1st batch | 250(?)    | 100       |311        | 
|nextToken          |Yes        |Yes         |NA         |
| success?          | **No**       |Yes        |Yes        |

Description if not succesful: stops after 250 data points, and graph does not show any more. See figure below, same response was showing for the aggregate test above. 

From Grafana: 

image

From Boto3:

image

#### Test 1.2.2.1 - single query
Usually works, but depending on the timeframe (Get different number of results by selecting '<' or '>' next to the timeframe). Sometimes it breaks even though property history value indicates there is data. Seems to be more unstable close to areas where datapoints are starting. (E.g. test case 2, where it shows 'No data' until a certain point when data points seem to be covering the majority of the timeframe.)

#### Test 1.2.2.1 - single query + expression

Same as previous version. Stops after 250 data points, and graph does not show any more. See similar problem in test 1.1.2.1

#### Test 1.2.3.1 - single query
Same as previous versions. 

#### Test 1.2.3.2 - single query + expressions

**Breaks and show 'No Data'**
Works when changing 'Format' to _time series_, but this is new compared to earlier versions. Does still break after 250 values. 

# Test 2 with available data points at the end desired timeframe
For this test, ensure that there exists data points from around half of the timeframe and later. Should also be more than 250 data points availble in the frame (more than maxResults requirement for the aws sitewise API).  Meaning it should be a subset like the figure shows below: 

image

#### Test 2.1.1.1 - single query
Aggregate resolution 'AUTO' in Grafana or '1m' in API, and duration of 630 minutes
|Checks             |   Grafana |    Boto3  |   AWS CLI |
|--                 |-----------|-----------|-----------|
|Total # points     |No Data        |260        |260        |
|# points 1st batch |No Data   | 0       |260        | 
|nextToken          |NA         |Yes         |NA         |
| success?          | **No**       |Yes        |Yes        |

Description of  failure: showing "No data", even though we know there exists data. Could it be because the first query gives no values in response, and somehow it does not check if there is a nextToken for values later in the timeframe?

### Test 2.1.2 - Grafana IoT SiteWise datasource version 1.5.0

Same result for version 1.5.0. See comment in Test 1.2.2.1 - single query

### Test 2.1.3 - Grafana IoT SiteWise datasource version 1.5.1

Same result as earlier. Show 'No Data'

### Test 2.2.1 - Grafana IoT SiteWise datasource version 1.4.1

Get Property Value History and duration of more than 250 minutes for our case (1 test point a minute maxResults from API is 250)

#### Test 2.2.1.1 - single query

Duration of 250 minutes
|Checks             |   Grafana |    Boto3  |   AWS CLI |
|--                 |-----------|-----------|-----------|
|Total # points     |269        |269        |269        |
|# points 1st batch |250(?)     |100        |269        | 
|nextToken          |Yes        |Yes        |NA        |
| success?          |Yes       |Yes        |Yes        |

image

#### Test 2.2.1.1 - single query + expression

|Checks             |   Grafana |    Boto3  |   AWS CLI |
|--                 |-----------|-----------|-----------|
|Total # points     |250        |269        |269        |
|# points 1st batch | 250(?)    | 100       |269        | 
|nextToken          |Yes        |Yes         |NA         |
| success?          | **No**       |Yes        |Yes        |

Description if not succesful: stops after 250 data points, and graph does not show any more. Show difference of response just by adding a math expression to the query ($A+1).

image

### Test 2.2.2 - Grafana IoT SiteWise datasource version 1.5.0
Same result for this version

### Test 2.2.3 - Grafana IoT SiteWise datasource version 1.5.1
Same result for this version

# Test 3 with available data points at the beginning desired timeframe

### Version 1.4.1, 1.5.0, 1.5.1

This test case results in the same as Test 1, having data points throughout the timeframe as wellas before and after. Works, but limits the datapoints when adding expressions for both aggregates and raw data (history value).

kevinwcyu:

This PR fixes an issue where queries with an expression do not show all data points.

Fixes #160, #182, #195

To be able to test this, you can use the Sitewise assets in the external dev account. Note, the demo assets live for 7 days before being deleted. If you see an error indicating that the resources can not be found, you can go to the Sitewise dashboard in the AWS console and create demo assets (there is a button to do so in the top right of the dashboard page)

The main issue here is that expression queries do not call the datasource's query method. Instead [if an expression query is present](https://github.com/grafana/grafana/blob/8235f774aa1afb49254861f3d6d3ea691baedde3/public/app/features/query/state/runRequest.ts#L203), the query is handled through the ExpressionDatasource's query handler. Since Sitewise paginates the aggregate and value history queries, we are no longer able to handle the pagination because the query is being executed by the ExecutionDatasources's query handler instead.

In the backend, we can see if a query is executed from the expression datasource by checking a header in the request. If it's a query execution we set the maximum number of datapoints/pages to receive to a high enough number so that we get back all of the data for the request time frame.

Here's a sample of a query that would not have worked before. Previously it would cut off after 250 data points.

Now it should show all of the data points. You can also change it to be a property value history as well.
sample query

Note, when running a property value history query with an expression, you'll need to set the format to `Time series`

### DIFF
--- a/pkg/server/handlers.go
+++ b/pkg/server/handlers.go
@@ -84,7 +84,8 @@ func (s *Server) handlePropertyValueHistoryQuery(ctx context.Context, req *backe
 	// Expressions need to run synchronously so we set MaxPageAggregations
 	// and MaxDataPoints to infinity to ensure that the query is not paginated.
 	_, isFromExpression := req.Headers["http_X-Grafana-From-Expr"]
-	if isFromExpression {
+	_, isFromAlert := req.Headers["FromAlert"]
+	if isFromAlert || isFromExpression {
 		query.MaxPageAggregations = int(math.Inf(1))
 		query.MaxDataPoints = int64(math.Inf(1))
 	}
@@ -119,7 +120,8 @@ func (s *Server) handlePropertyAggregateQuery(ctx context.Context, req *backend.
 	// Expressions need to run synchronously so we set MaxPageAggregations
 	// and MaxDataPoints to infinity to ensure that the query is not paginated.
 	_, isFromExpression := req.Headers["http_X-Grafana-From-Expr"]
-	if isFromExpression {
+	_, isFromAlert := req.Headers["FromAlert"]
+	if isFromAlert || isFromExpression {
 		query.MaxPageAggregations = int(math.Inf(1))
 		query.MaxDataPoints = int64(math.Inf(1))
 	}
--- a/pkg/server/handlers.go
+++ b/pkg/server/handlers.go
@@ -81,16 +81,24 @@ func (s *Server) handlePropertyValueHistoryQuery(ctx context.Context, req *backe
 		return DataResponseErrorUnmarshal(err)
 	}
 
+	// Expressions need to run synchronously so we set MaxPageAggregations
+	// and MaxDataPoints to infinity to ensure that the query is not paginated.
+	_, isFromExpression := req.Headers["http_X-Grafana-From-Expr"]
+	if isFromExpression {
+		query.MaxPageAggregations = int(math.Inf(1))
+		query.MaxDataPoints = int64(math.Inf(1))
+	}
+
 	frames, err := s.Datasource.HandleGetAssetPropertyValueHistoryQuery(ctx, query)
 	if err != nil {
 		return DataResponseErrorRequestFailed(err)
 	}
 
 	if len(frames) > 0 && query.ResponseFormat == "timeseries" {
-		for _, frame := range frames {
+		for i, frame := range frames {
 			wide, err := data.LongToWide(frame, &data.FillMissing{Mode: data.FillModeNull, Value: math.NaN()})
 			if err == nil {
-				frames = []*data.Frame{wide}
+				frames[i] = wide
 			}
 		}
 	}
@@ -106,6 +114,14 @@ func (s *Server) handlePropertyAggregateQuery(ctx context.Context, req *backend.
 	query, err := models.GetAssetPropertyValueQuery(&q)
 	if err != nil {
 		return DataResponseErrorUnmarshal(err)
+	}
+
+	// Expressions need to run synchronously so we set MaxPageAggregations
+	// and MaxDataPoints to infinity to ensure that the query is not paginated.
+	_, isFromExpression := req.Headers["http_X-Grafana-From-Expr"]
+	if isFromExpression {
+		query.MaxPageAggregations = int(math.Inf(1))
+		query.MaxDataPoints = int64(math.Inf(1))
 	}
 
 	frames, err := s.Datasource.HandleGetAssetPropertyAggregateQuery(ctx, query)
--- a/pkg/server/test/property_value_aggregate_test.go
+++ b/pkg/server/test/property_value_aggregate_test.go
@@ -2,6 +2,7 @@ package test
 
 import (
 	"context"
+	"math"
 	"testing"
 	"time"
 
@@ -34,8 +35,8 @@ func Test_property_value_aggregate_query_by_asset_id_and_property_id(t *testing.
 				*entries.PropertyId == "11propid-aaaa-2222-bbbb-3333cccc4444" &&
 				*entries.AggregateTypes[0] == "SUM"
 		}),
-		mock.Anything,
-		mock.Anything,
+		1,
+		0,
 	).Return(&iotsitewise.BatchGetAssetPropertyAggregatesOutput{
 		NextToken: Pointer("some-next-token"),
 		SuccessEntries: []*iotsitewise.BatchGetAssetPropertyAggregatesSuccessEntry{{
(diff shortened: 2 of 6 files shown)


pkg/server/test/property_value_aggregate_test.go
```diff
@@ -34,8 +35,91 @@ func Test_property_value_aggregate_query_by_asset_id_and_property_id(t *testing.
 				*entries.PropertyId == "11propid-aaaa-2222-bbbb-3333cccc4444" &&
 				*entries.AggregateTypes[0] == "SUM"
 		}),
+		1,
```
iwysiu, line 38: It might make sense to do this old test and the new one as [table tests](https://dave.cheney.net/2019/05/07/prefer-table-driven-tests), since they're almost the same and very wordy.

pkg/server/test/property_value_aggregate_test.go
```diff
@@ -121,8 +206,105 @@ func Test_property_value_aggregate_query_by_property_alias(t *testing.T) {
 				*entries.PropertyAlias == "/amazon/renton/1/rpm" &&
 				*entries.AggregateTypes[0] == "SUM"
 		}),
+		1,
```
iwysiu, line 209: same comment about the table tests

pkg/server/test/property_value_history_test.go
```diff
@@ -55,17 +60,19 @@ func Test_getPropertyValueHistoryHappyCaseTable(t *testing.T) {
 	require.Nil(t, err)
 
 	for i, dr := range qdr.Responses {
-		fname := fmt.Sprintf("%s-%s.golden", "property-history-values-table", i)
+		fname := fmt.Sprintf("%s-%s.golden", "property-history-values-boolean", i)
 		experimental.CheckGoldenJSONResponse(t, "../../testdata", fname, &dr, true)
 	}
 }
 
-func Test_getPropertyValueHistoryHappyCaseTimeSeries(t *testing.T) {
```
iwysiu, line 63: Did this test go missing?

pkg/server/test/property_value_history_test.go
```diff
@@ -18,9 +23,9 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-func Test_getPropertyValueHistoryHappyCaseTable(t *testing.T) {
```
iwysiu, line 21: Did this test go missing?

pkg/server/test/property_value_aggregate_test.go
fridgepoet: Just wanted to mention that though the individual tests are wordy, table tests come with their own disadvantage that they create a lot of jumping around to read a test from top to bottom "as a story."

After growing a test and copying/pasting, this can create a situation where the test input is hundreds of lines away from the function call under test and assertions.

iwysiu: That's fair. I would still lean toward table tests on these, just because these tests specifically have a lot of boilerplate due to the frames that makes it kind of hard to parse what's actually being tested.

pkg/server/test/property_value_history_test.go
kevinwcyu: i replaced this test with `Test_get_property_value_history_with_default_aka_table_response_format` which inlines the inputs and test outputs instead of loading separate json files. let me try to reorder to see if this improves the diffs here.

pkg/server/handlers.go
```diff
@@ -81,16 +81,24 @@ func (s *Server) handlePropertyValueHistoryQuery(ctx context.Context, req *backe
 		return DataResponseErrorUnmarshal(err)
 	}
 
+	// Expressions need to run synchronously so we set MaxPageAggregations
+	// and MaxDataPoints to infinity to ensure that the query is not paginated.
+	_, isFromExpression := req.Headers["http_X-Grafana-From-Expr"]
+	if isFromExpression {
+		query.MaxPageAggregations = int(math.Inf(1))
+		query.MaxDataPoints = int64(math.Inf(1))
```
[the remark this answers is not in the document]
kevinwcyu: the 20000 limit is per page, but we might actually have more than 20000 data points for the query.

i tried with -1 but we break out of the paginator based on the max data points `return numPages < maxPages && count <= maxResults` if that's false, then it stops fetching more data points. here's a go playground https://go.dev/play/p/aD2UD2HBWeO showing that we'd break out of the paginator after the first page of results was fetched.

pkg/server/handlers.go
```diff
@@ -81,16 +81,24 @@ func (s *Server) handlePropertyValueHistoryQuery(ctx context.Context, req *backe
 		return DataResponseErrorUnmarshal(err)
 	}
 
+	// Expressions need to run synchronously so we set MaxPageAggregations
+	// and MaxDataPoints to infinity to ensure that the query is not paginated.
+	_, isFromExpression := req.Headers["http_X-Grafana-From-Expr"]
```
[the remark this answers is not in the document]
kevinwcyu: yea, alerts will run on this query type too. i'll have to check how it works though, the demo data expired so will have to wait a bit for more data to test it.

pkg/sitewise/client/client.go
```diff
@@ -56,7 +56,17 @@ func (c *sitewiseClient) BatchGetAssetPropertyValueHistoryPageAggregation(ctx co
 		if len(output.SuccessEntries) > 0 {
 			count += len(output.SuccessEntries[0].AssetPropertyValueHistory)
 		}
-		success = append(success, output.SuccessEntries...)
+		if len(success) > 0 {
+			for i, entry := range success {
+				for _, successEntry := range output.SuccessEntries {
+					if *successEntry.EntryId == *entry.EntryId {
```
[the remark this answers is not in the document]
iwysiu: (do we know that the first call has all the entryids?)

iwysiu (commented):
One more question about the logic in client.go ,  but makes sense to me besides that.
If you end up needing to do the same thing for alerts, you should just be able to check for the alert header in the same place as the expression one

pkg/server/test/property_value_aggregate_test.go
kevinwcyu: will take a look at table tests for these today

iwysiu (approved):

PR #196, opened 2023-06-05
Remove code coverage and change compatibility check

fridgepoet:
Our squad has decided to stop using the code coverage report and to remove the breaking changes report. This PR addresses that.

This will:
* Remove the unit test code coverage changes report
* Tone down the breaking changes report so that it doesn't not add a comment and leaves the check in place. Originally there [was a concern about the Compatibility check which was always failing](https://github.com/grafana/athena-datasource/pull/195), but it doesn't seem to be the case now. So we'll move that out of comment and back to just a status check.


.github/workflows/detect-breaking-changes.yml
```diff
@@ -17,6 +17,6 @@ jobs:
         uses: grafana/plugin-actions/is-compatible@v1
         with:
           module: './src/module.ts'
-          comment-pr: 'yes'
-          fail-if-incompatible: 'no'
```
sarahzinger, line 21: Am I remembering correctly that there was an issue where levity checks always fail for external contributors? Maybe a reason to keep it as just a comment?

fridgepoet: Yeah that might still be the case. https://raintank-corp.slack.com/archives/C01C4K8DETW/p1680805987638169
Just trying to understand here: If we have an external contribution but keep it as a comment, what is the difference for you? (Is it that it's very bad that we ever merge anything that is failing a status check?)

sarahzinger: I don't really love having a status failure that we ignore, it feels like better to just remove the check all together? But I guess I just don't totally follow what "fail if incompatible" will do here, since it feels like we still get a status failure even if that's set to no?

I'm game to try this change though! We can always see what happens and try again later haha

fridgepoet: Okay I'm hoping it is not something we will ignore most of the time at least. Hope that this change is good enough to try for now, happy to make changes later again.

PR #203, opened 2023-07-19
Fix: Property aggregate queries returning duplicated data

### LINKED ISSUE #200: Property Value Aggregate Query Has Double the Expected Number of Data Points
The following description was taken from 2 comments in https://github.com/grafana/iot-sitewise-datasource/issues/160

---

I tested the new version (1.9.2) with Grafana version 10.0.0. 

With my current setup, I get the following response:

image

By downloading the data and looking at it, there are duplicates of most of the values. Without knowing, I suspect it is the paginated response that is duplicated. 

Using [Boto3 - get_asset_property_aggregates](https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/iotsitewise/client/get_asset_property_aggregates.html) I get the expected value of **600** data points.

As the screenshot shows above, the expression fails when using the *value aggregates* (top right corner), only showing 1 data point. By simply taking the expression query away, the result would be the same as top left corner (as expected).

Similarly, the expression does not get all the values for *value history* either. Again, I suspect that this is something with pagination/next token, not being able to show all the result.  

Please let me know if anything is unclear or you require more information.

_Originally posted by @egheie in https://github.com/grafana/iot-sitewise-datasource/issues/160#issuecomment-1592591464_

---

Hi @kevinwcyu, no problem. 

Here are the screenshots, starting **bottom left**, going clockwise. Note: every quadrant will have the same query, just the panel is different. (Either _Time series_, or _Stat_ with `Calculation` = `Count`). 

When doing this, I saw that the bottom left _Stat_ was set to **Get property value aggregates** and not **Get property value history**. See new screenshot at the bottom, sorry for that. 

**Screenshots of query editor**:
Bottom left
image

Top left
image

Top right
image

Bottom right
image

Updated screenshot:
image

_Originally posted by @egheie in https://github.com/grafana/iot-sitewise-datasource/issues/160#issuecomment-1594699586_

kevinwcyu:

Fixes an issue where some data points are duplicated.

Fixes #200

Here's a sample python script that can fetch the expected number of data points for a property aggregate query using boto3. Make sure you have boto3 installed. You may need to update the `assetId`, `propertyId`, `startDate`, and `endDate` if the sample data in external dev has expired and you recreate the demo data. Your Sitewise data source should be configured for the external dev account. You can use the Stat panel and set the calculation value to "Count" to compare the number of data points from the script below (just make sure the time range matches the python script).

```py
import boto3
from datetime import datetime

client = boto3.client('iotsitewise')

# get asset property aggregates
# https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/iotsitewise/client/get_asset_property_aggregates.html
agg_paginator = client.get_paginator('get_asset_property_aggregates')
agg_iterator = agg_paginator.paginate(
    assetId='fc1cbae5-0a28-4152-b622-35e92493d62e',
    propertyId='490c9933-b87e-4d51-9512-95e7ca476ed9',
    startDate=datetime.fromisoformat('2023-07-18T07:00:00Z'),
    endDate=datetime.fromisoformat('2023-07-19T06:59:59Z'),
    timeOrdering='ASCENDING',
    aggregateTypes=['AVERAGE'],
    resolution='1m',
    maxResults=250,
)
agg_count = 0
for p in agg_iterator:
  agg_count += len(p['aggregatedValues'])
print(agg_count)
```

### DIFF
--- a/pkg/sitewise/client/client.go
+++ b/pkg/sitewise/client/client.go
@@ -144,14 +144,6 @@ func (c *sitewiseClient) BatchGetAssetPropertyAggregatesPageAggregation(ctx cont
 					success = append(success, successEntry)
 				}
 			}
-
-			for i, entry := range success {
-				for _, successEntry := range output.SuccessEntries {
-					if *successEntry.EntryId == *entry.EntryId {
-						success[i].AggregatedValues = append(success[i].AggregatedValues, successEntry.AggregatedValues...)
-					}
-				}
-			}
 		} else {
 			success = append(success, output.SuccessEntries...)
 		}


pkg/sitewise/client/client.go
```diff
@@ -144,14 +144,6 @@ func (c *sitewiseClient) BatchGetAssetPropertyAggregatesPageAggregation(ctx cont
...
 				}
 			}
-
-			for i, entry := range success {
-				for _, successEntry := range output.SuccessEntries {
-					if *successEntry.EntryId == *entry.EntryId {
-						success[i].AggregatedValues = append(success[i].AggregatedValues, successEntry.AggregatedValues...)
-					}
-				}
-			}
```
kevinwcyu, line 154: This was a bad copy/paste, or rebase or something where this got merged. It re-appended the results that were already appended.

fridgepoet (approved):

PR #205, opened 2023-07-23
Fix scoped variables replacement in assetids such as repeat panels

### LINKED ISSUE #178: Failed to fetch query data when a template variable of assets was selected as multiple items or 'All'
**What happened**:

I have dashboards depending on template variables which store list of multiple assets of SiteWise for repeating panels.

For plugin version 1.7.0 or 1.8.0, if the variable's **Multi-value** or **All** option was enabled and I used them, panels did not render correct values due to the query data issue.

Same as behavior as plugin version 1.5.1 (see screenshots)

1. On AWS IoT Sitewise, create an asset model like `Example model` with any property like `foo`
1. Create two or more assets based on `Example model`, like `Asset-A` and `Asset-B`.
1. Edit those assets to set any value to `foo`.
1. On Grafana, go to any dashboard and create a variable like a `assets` using a SiteWise datasource, then select **List assets** in the Query type field and `Example model` in the **Model ID** field. Then check both of **Multi-value** and **Include All option** options, and click **Apply**.
1. Add a panel, then:
    * On **Query** tab:
        * Select a SiteWise datasource
        * Select **Get property value** as Query type
        * Select `${assets}` as Asset
        * Select `foo` as Property
    * Select **Stat** as visualization:
        * Select `assets` as **Repeat by variable**
        * Select `foo` as **Fields** in **Value Options** section
    * Click **Apply** and close the panel.
1. Click `assets` variable dropdown on the dashboard:
    * Select only `Asset-A` -> The panel was rendered as expected.
    * Select `Asset-A` and `Asset-B` -> Two panels were rendered, but values of them are not correct (shown as 'No data').
    * Select **All** -> Panels show error as attached screenshot.
 
**Screenshots**

Plugin version 1.8.0:

* Single item was selected:
    * v1 8 0-select-a-value
* Two items were selected:
    * v1 8 0-select-two-values
* 'All' was selected:
    * v1 8 0_select-All

Plugin version 1.5.1 (expected behavior)

* Single item was selected:
    * v1 5 1_select-a-value
* Two items were selected:
    * v1 5 1_select-two-values
* 'All' was selected:
    * v1 5 1_select-All

**Environment**:
- Grafana version: 9.4.7 or 9.5.1
- Plugin version: 1.7.0 or 1.8.0 (1.5.1 is OK)
- OS Grafana is installed on: Ubuntu 22.04
- User OS & Browser: Windows 10 & Chrome
- Others:

Thanks in advance,

ahom:
This PR allows users to create repeat panels or rows when iterating over multiple assetIds. The repeat functionality was completely broken when trying to use it with the assetId property because ScopedVariables was not used during the replacement.

Fixes #178 

Previous implementation of the function was not taking into account the scopedVars that are needed for the repeat functionality within Grafana. I took the liberty to simplify the implementation, leveraging the 'csv' formatter in order to gracefully handle all cases (constant value, single-value and muti-value vars). This code won't work only if an assetId includes a comma, which I don't believe is valid.
I also removed the previous unit tests which are not relevant anymore as they were testing the function body with the mock of TemplateSrv. Happy to add any relevant test if there is anything that would be meaningful.


idastambuk:
Hi @ahom, thanks for the contribution! 
I tested it with repeating variables it looks good now. 
However, I wouldn't delete the old tests, as they do test previous bugs (just modify them, but keep the test cases). I think an additional one should be added for this fix as well, for scopedVars. Let us know if you need help with that.

ahom:
Thanks @idastambuk for the comments!
I added back the tests but had to modify the way they were working, I also added two new tests for ScopedVars as suggested.

idastambuk:
Hi @ahom sorry for the delay! Looks good, will merge this and release it

idastambuk (approved):

PR #206, opened 2023-07-24
Fix: Query with expression only returns partial data

### LINKED ISSUE #202: Query with expression only returns partial data
I could reproduce it with the demo data. See gif below for the result.  So you had no issues with the expressions and both time series (query + expression) cover the entire expected time range? 

For all panels, the following was used:
Asset used: _Demo Turbine Asset 1_
Property: _Wind Speed_

grafana-debug

_Originally posted by @egheie in https://github.com/grafana/iot-sitewise-datasource/issues/160#issuecomment-1600397788_

---

Another short gif to show how it changes with the expression being disabled or not and by removing the expression completely. 

There is also something a bit funky happening with the name in the legend in the bottom right corner when adding an expression. Raised a similar issue in #198, but worth mentioning here as well. 

grafana-debug-2

_Originally posted by @egheie in https://github.com/grafana/iot-sitewise-datasource/issues/160#issuecomment-1600432894_

kevinwcyu:

Fixes an issue where queries with expression return partial data only. In a previous PR queries with expressions were set to execute completely on the backend before returning data to the frontend. This was done by setting the maximum number of datapoints and pages to fetch to `math.Inf(1)`. This worked locally, but when the Sitewise data source was built, `math.Inf(1)` ended up being a negative number. This caused the backend to only return the first page of data because the function that fetches the data from Sitewise would check if the current number of datapoints/pages is less than the maximum number of datapoints/pages it should fetch https://github.com/grafana/iot-sitewise-datasource/blob/c70aadea564390be9388ba2d8eaa11c273229fa3/pkg/sitewise/client/client.go#L79

Since the number is negative, it just ends up always fetching only 1 page of data before returning. Since the expression data source doesn't support paged queries, it would not follow the next token to continue fetching the remaining data.

Fixes #202 

To test this, it's better to download the drone build of the data source instead of pulling it and running it locally. To test this, you need to make sure there is enough sample data to cause the query to need to be paged (I think it's about 4.5 hours at least if doing a property aggregate query using a 1m resolution).

I changed the max data points/pages from `math.Inf(1)` to `math.MaxInt32`, but let me know if you think there's a better number we should set it to.

### DIFF
--- a/pkg/server/handlers.go
+++ b/pkg/server/handlers.go
@@ -86,8 +86,8 @@ func (s *Server) handlePropertyValueHistoryQuery(ctx context.Context, req *backe
 	_, isFromExpression := req.Headers["http_X-Grafana-From-Expr"]
 	_, isFromAlert := req.Headers["FromAlert"]
 	if isFromAlert || isFromExpression {
-		query.MaxPageAggregations = int(math.Inf(1))
-		query.MaxDataPoints = int64(math.Inf(1))
+		query.MaxPageAggregations = math.MaxInt32
+		query.MaxDataPoints = math.MaxInt32
 	}
 
 	frames, err := s.Datasource.HandleGetAssetPropertyValueHistoryQuery(ctx, query)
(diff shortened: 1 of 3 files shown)


fridgepoet (approved):
Wow! What a mess...
It's this issue? https://github.com/golang/go/issues/56023

Issue #237, opened 2023-10-24
Disassociated Data Stream alias cannot be longer than 64 characters


alexhegg-aws:
**What happened**:
Using a disassociated data stream where the alias is longer than 64 characters results in an error

**What you expected to happen**:
Data should be returned. Associating the stream with an asset allows it to be queried.

**How to reproduce it (as minimally and precisely as possible)**:
Create 2 disassociated data streams named:
/123456789/123456789/123456789/123456789/123456789/123456789/s
/123456789/123456789/123456789/123456789/123456789/123456789/long

Create 2 queries of type "Get property history value"

Enter both property aliases. The first (62 characters) returns data, the second returns an eror:
failed to fetch query data: InvalidRequestException: 1 validation error detected: Value '_123456789_123456789_123456789_123456789_123456789_123456789_long' at 'entries.1.member.entryId' failed to satisfy constraint: Member must have length less than or equal to 64

**Screenshots**
![image](https://github.com/grafana/iot-sitewise-datasource/assets/115964162/3566f69d-2913-44b0-b3bd-f5daf92ddaf8)

**Anything else we need to know?**:

**Environment**:
- Grafana version: 10.1.5
- Plugin version: 1.11.1
- OS Grafana is installed on: Linux
- User OS & Browser: MacOS + Chrome
- Others:

PR #239, opened 2023-10-25
Disassociated streams: Hash entryId to fix bug with property aliases longer than 64 characters

idastambuk:
When implementing disassociated data streams, I set the entryID to be equal to the propertyAlias. Normally for associated streams it is set to assetID. This isn't possible for disassociated streams since they don't have an asset connected to them. 
However, I didn't account for property aliases that are longer than 64 characters. 
One possible solution is to hash the property aliases. As far as I can see, they don't need to be in human readable format, since entryId is used on our BE to:
1. send along with queries to distinguish them and
2. access the results of those queries, be they success or errors, and append them to our frames. 

The names of the frames/fields don't come from entryIDs but from the property aliases from the query, which are readable. 

Another solution would be to just shorten the string, but it is possible to have 2 property aliases that are almost the same apart from the end. This would create two equal entryIDs, so I didn't go with this approach.

Fixes https://github.com/grafana/iot-sitewise-datasource/issues/237

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 All notable changes to this project will be documented in this file.
 
+## v1.12.1
+- Disassociated streams: Hash entryId to fix bug with property aliases longer than 64 characters in [#239](https://github.com/grafana/iot-sitewise-datasource/pull/239)
+
 ## v1.12.0
 - Query by property alias: Add support for unassociated streams in [#231](https://github.com/grafana/iot-sitewise-datasource/pull/231)
 
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "grafana-iot-sitewise-datasource",
-  "version": "1.12.0",
+  "version": "1.12.1",
   "description": "View IoT Sitewise data in grafana",
   "scripts": {
     "build": "webpack -c ./.config/webpack/webpack.config.ts --env production",
--- a/pkg/resource/query_resouce_provider.go
+++ b/pkg/resource/query_resouce_provider.go
@@ -59,7 +59,7 @@ func (rp *queryResourceProvider) Properties(ctx context.Context) (map[string]*io
 	properties := map[string]*iotsitewise.DescribeAssetPropertyOutput{}
 	// if the query for a PropertyAlias doesn't have an assetId or propertyId, it means it's a disassociated stream
 	// in that case, we call Property() with empty values, which will set AssetProperty.Name to the alias
-	// and will set the EntryId to the alias (to access values in results)
+	// and will set the EntryId to the hashed alias (to access values in results)
 	if len(rp.baseQuery.AssetIds) == 0 && rp.baseQuery.PropertyId == "" && rp.baseQuery.PropertyAlias != "" {
 		prop, err := rp.resources.Property(ctx, "", "", rp.baseQuery.PropertyAlias)
 		if err != nil {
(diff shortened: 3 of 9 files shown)


jhoulihan:
Greetings, I firstly admit, I don't fully understand the depth of technical issues.  That aside, I was triggered immediately by the assumption of xx not being human readable.

My use case is that our policy is to develop grafana pages in dev, migrate to test, migrate to prod.  
The comment maybe 1000% unrelated, but from my context, using UUID's (eg sitewise models and assets) / hashes (as proposed ) being non-human readable completely breaks the workflow given that AWS/AMG dev account, AWS/AMG test account, AWS/AMG prod account produces different UUID's and I'm assuming hashes presenting a similar problem.

Datastreams are the facts, the measurements and the truth. 

Models and assets that contain datastreams are just metadata , human constructs and context of the truth (datastreams)

Please consider to preserve the datastream legibility because they are _the data_ asset and priority.

idastambuk:
> Greetings, I firstly admit, I don't fully understand the depth of technical issues. That aside, I was triggered immediately by the assumption of xx not being human readable.
> 
> My use case is that our policy is to develop grafana pages in dev, migrate to test, migrate to prod. The comment maybe 1000% unrelated, but from my context, using UUID's (eg sitewise models and assets) / hashes (as proposed ) being non-human readable completely breaks the workflow given that AWS/AMG dev account, AWS/AMG test account, AWS/AMG prod account produces different UUID's and I'm assuming hashes presenting a similar problem.
> 
> Datastreams are the facts, the measurements and the truth.
> 
> Models and assets that contain datastreams are just metadata , human constructs and context of the truth (datastreams)
> 
> Please consider to preserve the datastream legibility because they are _the data_ asset and priority.

Hi @jhoulihan can you elaborate on "preserving the datastream legibility"? Can you explain in which cases it would be important to preserve the human readable format? 
Hashes and UUID's are not stored in Grafana in panels querying disassociated streams. The only thing that is needed is the propertyAlias which we use to query Sitewise. EntryId here, as explained in the description, is a temporary string that loses relevance for Grafana as soon as we process the data frame.
I admit I don't know the context of how AMG migrates dashboards, but of course this temporary hash could be relevant for the Sitewise service and logs itself, or AMG's migrations themselves. I would like to understand the use case in order to evaluate a possible alternative solution. Thanks!

jhoulihan:
Thanls Ida,
For example..  I can simply paste in the property alias/sitewise datastream that is disassociated from any asset
image

This is legible to promote from dev/test/prod

Using sitewise asset approach to create a query, UUID's are generated that break being able to migrate grafana dashboards from dev/test/prod as the environments are housed in different AWS accounts - therefore UUID's differ between accounts and unable to identify what datastream the UUID is associated with , without significant development effort.
image

I'm treating the data ingestion, storage and presentation of datastreams as an enterprise solution. In that context, it means that I need to consider and support the sitewise/grafana stack with people who are not software engineers who can decipher UUIDs, hashes and similar constructs.  They need a simple instruction to copy/paste the json from one account to another (ie dev/test/prod) and only have to find/replace the UUID of the datasource for each account (eventhough that can be improved on - but leave that to another day)

I hope this helps ? 
Thanks again Ida

idastambuk:
Hi @jhoulihan thanks for clarifying. In the context of unassociated streams that are queried only by propertyId, the hash not being legible will not be an issue, because these hashed are not stored in Grafana in panels querying disassociated streams. The only thing that is stored is the propertyAlias which we use to query Sitewise, which makes the json as readable as the propertyAlias (and which is the responsibility of the user). For such panels migration will be trivial, as you mention above. 

As for propertyIDs being stored as UUIDs in Grafana panel JSON, this is the case with data streams that are queried with propertyId, and I'm not sure anything can be done about it, except for the user to add an alias. We do support querying streams with just an alias. This way the user's panel doesn't have to reference any UUIDs.

I hope that clears it up! 
In conclusion, as the hashes never reach panel JSON, I think we're good to proceed with merging this.

PR #244, opened 2023-11-06
Query and Config editors: Migrate to new form styling under feature toggle 

idastambuk:
Migrating Config Editor to the new design system form components, as well as the Query Editor to use EditorField components from grafana/experimental.
The new components are under a Grafana feature flag `awsDatasourcesNewFormStyling`.

### Config Editor
**What was done**
Replace the editor inline fields for the grafana/ui Field components if the feature flag is enabled

Screenshot 2023-11-06 at 2 13 39 PM

Screenshot 2023-11-06 at 2 13 17 PM

### Query Editor
**What was done**

- Use new vertical form fields instead of inline
- separated out the query options for Expand Time Range, Format, Quality, Time into a collapse. 
I did this because they were squeezed after other fields, like Property, and weren't that visible either. But if there's a strong reason for these being accessible right after the fields, I can put them back.
Screenshot 2023-11-06 at 11 53 07 AM
Screenshot 2023-11-06 at 2 46 27 PM

Fixes https://github.com/grafana/oss-plugin-partnerships/issues/634

### DIFF
--- a/jest-setup.js
+++ b/jest-setup.js
@@ -1,6 +1,6 @@
 // Jest setup provided by Grafana scaffolding
 import './.config/jest-setup';
-// Used by useMeasure
+// Used by LinkButton -> Text component from grafana/ui
 global.ResizeObserver = class ResizeObserver {
     //callback: ResizeObserverCallback;
   
--- a/jest-setup.js
+++ b/jest-setup.js
@@ -1,2 +1,32 @@
 // Jest setup provided by Grafana scaffolding
 import './.config/jest-setup';
+// Used by useMeasure
+global.ResizeObserver = class ResizeObserver {
+    //callback: ResizeObserverCallback;
+  
+    constructor(callback) {
+      setTimeout(() => {
+        callback(
+          [
+            {
+              contentRect: {
+                x: 1,
+                y: 2,
+                width: 500,
+                height: 500,
+                top: 100,
+                bottom: 0,
+                left: 100,
+                right: 0,
+              },
+              target: {},
+            },
+          ],
+          this
+        );
+      });
+    }
+    observe() {}
+    disconnect() {}
+    unobserve() {}
+  }
--- a/src/components/ConfigEditor.tsx
+++ b/src/components/ConfigEditor.tsx
@@ -96,16 +96,20 @@ export function ConfigEditor(props: Props) {
             <ConnectionConfig {...props} standardRegions={standardRegions} newFormStylingEnabled={true} />
           )}
 
-<Divider/>
+          <Divider />
           <ConfigSection title="Edge settings" data-testid="edge-settings">
             {!endpoint && (
-              <Alert title="Edge region requires an explicit endpoint configured above" severity="warning" data-testid="endpoint-warning" />
+              <Alert
+                title="Edge region requires an explicit endpoint configured above"
+                severity="warning"
+                data-testid="endpoint-warning"
+              />
             )}
 
             <Field label="Authentication Mode" htmlFor="edgeAuthMethods">
               <Select
-              id="edgeAuthMethods"
-              aria-label="Authentication Mode"
+                id="edgeAuthMethods"
+                aria-label="Authentication Mode"
                 options={edgeAuthMethods}
                 value={edgeAuthMode}
                 onChange={(v) => {
@@ -117,7 +121,7 @@ export function ConfigEditor(props: Props) {
               <>
                 <Field label="Username" description="The username set to local authentication proxy" htmlFor="username">
                   <Input
-                  id="username"
+                    id="username"
                     name="username"
                     value={jsonData.edgeAuthUser}
                     autoComplete="off"
@@ -126,9 +130,13 @@ export function ConfigEditor(props: Props) {
                     required
                   />
                 </Field>
-                <Field label="Password" description="The password sent to local authentication proxy" htmlFor="password">
+                <Field
+                  label="Password"
+                  description="The password sent to local authentication proxy"
+                  htmlFor="password"
+                >
                   <Input
-                  id="password"
+                    id="password"
                     type="password"
                     name="password"
                     autoComplete="off"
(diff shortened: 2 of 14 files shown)


jest-setup.js
```diff
@@ -1,2 +1,32 @@
 // Jest setup provided by Grafana scaffolding
 import './.config/jest-setup';
+// Used by useMeasure
```
kevinwcyu, line 3: is this in some library code? i couldn't find any usage of `useMeasure` inside sitewise

kevinwcyu (commented):
Before selecting an asset, there's an empty row. let's hide rendering that until the necessary information is required to render what should go in that row.

Screenshot 2023-11-06 at 11 18 17 AM

jest-setup.js
idastambuk: Copied it from grafana main and forgot to add the proper comment, smh. Fixed!

idastambuk:
> Before selecting an asset, there's an empty row. let's hide rendering that until the necessary information is required to render what should go in that row.
> 
> Screenshot 2023-11-06 at 11 18 17 AM

Great catch, just committed a change in logic that uses a new flag for the new form options row

src/types.ts
```diff
@@ -187,6 +187,15 @@ export function shouldShowLastObserved(queryType?: QueryType): boolean {
   return queryType === QueryType.PropertyAggregate || queryType === QueryType.PropertyValueHistory;
 }
 
+export function shouldShowOptionsRow(query: SitewiseQuery, showProp: boolean): boolean {
+  const shouldShowLastObservedSwitch =
+    shouldShowLastObserved(query.queryType) && !Boolean(query.propertyAlias) && showProp;
+  const shouldShowWithPropertyAlias =
+    // shouldn't show the row when querying associated assets with property alias, otherwise show it every time property alias is set
+    query.propertyAlias && !isListAssociatedAssetsQuery(query);
```
idastambuk, line 195: SO I changed the logic (and the UI) a little bit from what we currently have in main. Right now, the Quality, Time and Format rows are not displayed if the query is `queryPropertyValue` and the `propertyAlias` is used. It looks to me like this isn't intended, as this query can also return a time series that gets formatted by all these fields, as seen here: 
https://github.com/grafana/iot-sitewise-datasource/blob/main/pkg/framer/property_value.go#L23

Furthermore, there is an unintended side-effect in the old form where if you select an asset, then add a propertyAlias, the Asset field is hidden, but the formatting options stay because the assetId doesn't get cleared from the query object.
In any case, I think explicitly allowing these fields to be selected with propertyAlias query is the right way to go.

src/components/query/PropertyQueryEditor.tsx
```diff
@@ -258,6 +314,35 @@ export class PropertyQueryEditor extends PureComponent<Props, State> {
     );
   }
 
+  renderOptions(query: SitewiseQuery, showProp: boolean, showQuality: boolean) {
```
sarahzinger, line 317: maybe let's break this up into a functional component rather than building another additional render function on this class? I think smaller functional components tend to be a bit easier to read and can even have some performance boosts

src/components/query/QueryEditor.tsx
```diff
@@ -88,16 +123,16 @@ export class QueryEditor extends PureComponent<Props> {
...
               menuPlacement="bottom"
             />
           </InlineField>
         </div>
-        {this.renderQuery(query)}
-      </>
-    );
-  }
+      
+      {renderQuery(query)}</>}
```
sarahzinger, line 135: Let's take this one more step and make this a functional react component so we can write:
```
<Query query={query} />
```

Or ideally some more descriptive name than that haha

sarahzinger (commented):
So much good work here! I'm curious if you think we can break this up into smaller prs at all? For example I see some good work here to refactor from a class component to a functional component, could that be done in a separate pr first, then we move to the styling? Or maybe tests in their own pr? Or the logic change in types.ts, is that directly tied to this new form styling? Curious what your thoughts are as I haven't dived into it too much!

src/components/ConfigEditor.tsx
```diff
@@ -65,136 +63,259 @@ export class ConfigEditor extends PureComponent<Props> {
...
-                className="width-30"
-                value={regions.find((region) => region.value === options.jsonData.defaultRegion)}
-                options={regions}
-                defaultValue={options.jsonData.defaultRegion}
-                allowCustomValue={true}
-                onChange={onUpdateDatasourceJsonDataOptionSelect(this.props, 'defaultRegion')}
-                formatCreateLabel={(r) => `Use region: ${r}`}
+    if (newFormStylingEnabled) {
+      return (
+        <div className="width-30">
```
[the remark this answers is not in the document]
idastambuk: I think it's ok right now, as this is a fixed class that implements grafana-specific spacing from grafana/ui. The only change I made for the new styling is I set it here once and not for each Field in props

idastambuk:
> So much good work here! I'm curious if you think we can break this up into smaller prs at all? For example I see some good work here to refactor from a class component to a functional component, could that be done in a separate pr first, then we move to the styling? Or maybe tests in their own pr? Or the logic change in types.ts, is that directly tied to this new form styling? Curious what your thoughts are as I haven't dived into it too much!

Most of the changes are tied to the restyling, so it won't be easy to separate them. The only one that isn't, is changing the component to a functional one, but that's just one component and I thought it would be an easy win (and not really a lot of code changes) so I slipped it in this one.
A possible solution to make this easier to review is to separate the restyling into Config and Query editor PRs, do you think it's worth it at this point?

src/components/query/QueryEditor.tsx
idastambuk: This is an old component, so if that's ok, I'll skip this to limit refactoring in this PR, as you mentioned above

src/components/ConfigEditor.tsx
```diff
@@ -20,28 +22,50 @@ const edgeAuthMethods: Array<SelectableValue<string>> = [
...
 
-export class ConfigEditor extends PureComponent<Props> {
-  constructor(props: Props) {
-    super(props);
-    this.state = {};
+export function ConfigEditor(props: Props) {
+  const newFormStylingEnabled = config.featureToggles.awsDatasourcesNewFormStyling;
+
+
+  // Simple
```
sarahzinger, line 29: What does this comment mean?

sarahzinger (approved):

src/components/ConfigEditor.tsx
idastambuk: Removed it!

PR #249, opened 2023-11-08
Upgrade yaml package

fridgepoet:
This PR upgrades cspell, fork-ts-checker-webpack-plugin and add cosmiconfig resolution in order to upgrade the package `yaml` version. The main reason the older version of `yaml` was used was due to a transitive dependency of an older version of `cosmiconfig`.

`fork-ts-checker-webpack-plugin` and `cspell` were on older versions requiring an older version of `cosmiconfig`.
`fork-ts-checker-webpack-plugin` and `cspell` were updated, but finally `@emotion/css` still used an older version of `cosmiconfig`. So the latest version of `cosmiconfig` was added to `resolutions`. 

Fixes https://github.com/grafana/oss-plugin-partnerships/issues/465

Is the following message during `yarn install` any issue?
```
Resolving packages...
warning Resolution field "cosmiconfig@8.3.6" is incompatible with requested version "cosmiconfig@^7.0.0"
warning Resolution field "cosmiconfig@8.3.6" is incompatible with requested version "cosmiconfig@8.0.0"
```

I sanity checked this locally and saw no issues.

Thank you Kevin and Ida for the help!

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,7 +46,7 @@
     "@typescript-eslint/eslint-plugin": "^5.42.1",
     "@typescript-eslint/parser": "^5.42.1",
     "copy-webpack-plugin": "^11.0.0",
-    "cspell": "6.13.3",
+    "cspell": "8.0.0",
     "css-loader": "^6.7.3",
     "eslint": "8.26.0",
     "eslint-config-prettier": "^8.3.0",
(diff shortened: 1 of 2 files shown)


sarahzinger (approved):
"Is the following message during yarn install any issue?" 

I think my guess is it could be but its not necessarily an issue? I think if you can run it locally and seems to build fine it's probably fine?

fridgepoet:
Sorry apparently a higher version of yaml was needed https://github.com/grafana/iot-sitewise-datasource/pull/253

PR #256, opened 2023-11-27
Support multiple assets for interpolated queries

### LINKED ISSUE #235: Bug: Interpolated query does not support multiple assets
**What happened**:
- Interpolated property values query does not support multiple assets

- Screenshot below should have separate lines for each asset

- The demo turbine data in SiteWise is enough, screenshot has all the query params you need.

**Screenshots**

Screenshot

**Environment**: Amazon Managed Grafana
- Grafana version: 9.4
- Plugin version: 1.10.1, and 1.11
- OS Grafana is installed on:
- User OS & Browser:
- Others:

kevinwcyu:

It allows selecting multiple assets when using an interpolated property values query.

Before:
before

After:
after

Fixes #235 

- As a part of this another bug was fixed where when multiple assets were selected and the property values history query was selected, and the time range was long enough where a multiple queries were needed to fetch all of the data points, it would error

Before:
2-before

After:
2-after

See the screenshots for the queries used for testing. Demo assets may need to be recreated in the external dev account if it expires.

### DIFF
--- a/cspell.config.json
+++ b/cspell.config.json
@@ -59,6 +59,9 @@
     "httptrace",
     "otelhttptrace",
     "idastambuk",
-    "fridgepoet"
+    "fridgepoet",
+    "errgroup",
+    "ectx",
+    "reqs"
   ]
 }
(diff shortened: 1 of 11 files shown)


idastambuk (approved):
Nice, looks good!

fridgepoet (approved):
🎉

PR #258, opened 2023-11-30
Use query region to get client for queries

### LINKED ISSUE #250: Switching region in query editor does not load assets
**What happened**:
When I switched in a query editor to a different region from Default, the assets present in that region didn't load.

In the query editor, switching regions loads that region's assets.

1. Create a datasource connection and set a Default region.
2. Go to Build a dashboard or to Explore
3. Select some query type, and click on "Select an asset" dropdown 
4. Now select a different region from the dropdown
5. Clicking on "Select an asset" dropdown does not load the different region's dropdown

Here is a screen recording. In this example I have set "us-west-2" as a Default region and have no assets there. I switch to "us-east-2" where I do have assets, but they do not load, even when I go to Explore.

https://github.com/grafana/iot-sitewise-datasource/assets/4163034/82e084b2-4975-4392-b906-c144e1a770c4

* Saving the new target asset's region in the data source configuration page, then going to dashboard/Explore does work
* Manually inputting the Asset ID into the query editor still does not load any data

**Environment**:
- Grafana version: Grafana v10.3.0-63184 (5285e9503be5702680acb2b52a6bda0632f4603d)
- Plugin version: 1.12.1
- OS Grafana is installed on: using squad Grafana Cloud instance
- User OS & Browser: MacOS and Chrome

iwysiu:
We weren't using the region from the query to get the client, despite sending it. This PR uses the region in the config sent to the SessionCache.

As a side note it feels like we have a lot of unnecessary abstraction in the sitewise plugin.

Fixes #250 

### DIFF
--- a/pkg/sitewise/client/client.go
+++ b/pkg/sitewise/client/client.go
@@ -168,7 +168,9 @@ func (c *sitewiseClient) BatchGetAssetPropertyAggregatesPageAggregation(ctx cont
 type AmazonSessionProvider func(c awsds.SessionConfig) (*session.Session, error)
 
 func GetClient(region string, settings models.AWSSiteWiseDataSourceSetting, provider AmazonSessionProvider) (client SitewiseClient, err error) {
-	sess, err := provider(awsds.SessionConfig{Settings: settings.ToAWSDatasourceSettings()})
+	awsSettings := settings.ToAWSDatasourceSettings()
+	awsSettings.Region = region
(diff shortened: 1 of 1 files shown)


fridgepoet (approved):
Great, works for me locally!

I'm sure one day when this code is being refactored, someone could really appreciate what seems like an obvious test today!

sarahzinger (approved):

PR #261, opened 2023-12-11
feat: implement all option for list associated assets query

mitchlee-amzn:
This pull request implements a flag to add an "**ALL**" option to the ListAssociatedAssets query. Listing children in this way currently isn't supported natively by the SiteWise API, so we work around this by first getting all the asset hierarchies, and requesting them individually, and unifying them at the end. This is temporary, as this API change will be added on the SiteWise end next year to improve scalability. It essentially enables the user to setup navigation through their asset hierarchies in the same way as the SiteWise Console Asset explorer does (and this uses the same approach as the console).

This is a critical feature request for one of our customers.

Fixes #

### DIFF
--- a/pkg/sitewise/api/list_associated_assets.go
+++ b/pkg/sitewise/api/list_associated_assets.go
@@ -17,29 +17,73 @@ func ListAssociatedAssets(ctx context.Context, client client.SitewiseClient, que
 	var (
 		hierarchyId        *string
 		traversalDirection *string
+		assetId            *string = util.GetAssetId(query.BaseQuery)
+		results            []*iotsitewise.AssociatedAssetsSummary
 	)
 
-	if query.HierarchyId != "" {
-		hierarchyId = aws.String(query.HierarchyId)
-		traversalDirection = aws.String("CHILD")
+	// Recursively load children
+	if query.LoadAllChildren {
+		asset, err := client.DescribeAsset(&iotsitewise.DescribeAssetInput{
+			AssetId: assetId,
+		})
+
+		if err != nil {
+			return nil, err
+		}
+
+		for _, h := range asset.AssetHierarchies {
+			// For this code path, we need to handle this internally, since it's a union of multiple queries
+			var nextToken *string = nil
+
+			for {
+				resp, err := client.ListAssociatedAssetsWithContext(ctx, &iotsitewise.ListAssociatedAssetsInput{
+					AssetId:     assetId,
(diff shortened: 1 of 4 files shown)


pkg/sitewise/api/list_associated_assets.go
```diff
@@ -14,6 +14,33 @@ import (
...
+		})
+
+		if err != nil {
+			return nil, err
+		}
+
+		var results []*iotsitewise.AssociatedAssetsSummary
+
+		for _, h := range asset.AssetHierarchies {
+			resp, _ := client.ListAssociatedAssets(&iotsitewise.ListAssociatedAssetsInput{
```
jiajzhou, line 32: 1.  you should handle error
2. the response may be paginated.

pkg/sitewise/api/list_associated_assets.go
```diff
@@ -14,6 +14,33 @@ import (
 
 func ListAssociatedAssets(ctx context.Context, client client.SitewiseClient, query models.ListAssociatedAssetsQuery) (*framer.AssociatedAssets, error) {
 
+	var assetId *string = util.GetAssetId(query.BaseQuery)
```
jiajzhou, line 17: [style] better follow the existing coding style, i.e. make the var declaration at the top of the func and then start implementation.

pkg/sitewise/api/list_associated_assets.go
```diff
@@ -14,6 +14,33 @@ import (
...
+
+	// Recursively load children
+	if query.LoadAllChildren {
```
jiajzhou, line 20: [style] I think it is better to be implemented together with the if statement on line 49, instead of a if check and return here.

jiajzhou (changes requested):
see the inline comments.

pkg/sitewise/api/list_associated_assets.go
mitchlee-amzn: This should be resolved, I also noticed I missed the maxResults and ctx parameters, so that's been added as well

pkg/sitewise/api/list_associated_assets.go
mitchlee-amzn: I think this is the way you suggest now, let me know if I misunderstood.

jiajzhou (approved):
LGTM

sarahzinger (approved):
Makes sense to me!

PR #262, opened 2023-12-14
fix(data-loading): concurrent fetch and caching of data

mitchlee-amzn:

Fixes #


pkg/sitewise/client/client.go
```diff
@@ -41,23 +42,139 @@ func NewSitewiseClientForRegion(region string) SitewiseClient {
...
+			intervalEnd := intervalStart + intervalDuration
+			if intervalEnd > endTime {
+				intervalEnd = endTime
+			}
+
+			data, err := fetchRawData(ctx, intervalStart, intervalEnd, c, req)
+			if err != nil {
+				fmt.Println(err.Error())
+				return
+			}
```
njvrzm, line 156: This introduces a significant change in behavior: Currently if any of its paginated requests receives an error, `BatchGetAssetPropertyValueHistoryPageAggregation` returns the error and no data. With this change, all request errors will be ignored and partial data may be returned. I'm not sure how likely this is in practice but it's certainly possible and it could lead to strange behavior.

Handling errors gracefully with parallel requests isn't necessarily simple, but to maintain consistency with the existing behavior, and the behavior of the other methods, it might be better to have the call to `BatchGetAssetPropertyValueHistoryPageAggregation` return no data and an error if any of the parallel calls encounters an error.

pkg/sitewise/client/client.go
```diff
@@ -41,23 +42,139 @@ func NewSitewiseClientForRegion(region string) SitewiseClient {
 	}
 }
 
+type IndexedData struct {
+	Index int
+	Data  *iotsitewise.BatchGetAssetPropertyValueHistoryOutput
+}
```
njvrzm, line 48: Very minor, but this type and its members might as well be lowercased since it's only of internal use.

njvrzm (changes requested):
Hello! I'm new on the AWS data sources team but had some time and thought I'd jump in here, hope it's helpful :)

Mostly minor comments, but two potentially significant ones - one regarding efficient result aggregation and one about a behavior change with respect to errors.

sarahzinger:
Hey Folks, 

thanks for taking a look at this, and bringing these performance issues to our attention! I think concurrent fetching and caching of data might add a fair bit of complexity to our system and might be a fairly large change with potentially unintended consequences. 

I'm wondering if we can first see if this [pr](https://github.com/grafana/iot-sitewise-datasource/pull/272) helps improve performance enough so that it's not an issue. We also are currently working to make this plugin streaming-capable the way twinmaker is, which I think might also be a better fit for this user. Perhaps we can revisit after those two prs are merged and if we still see performance issues we then consider adding in a frontend cache?

PR #274, opened 2024-01-11
Query Editor: Stop running queries on every change

idastambuk:
A proposal for a change in behavior of the query editor. 

Currently we call `onRunQuery()` on every change in the editor, even if the query isn't complete. This sometimes results in errors from the backend/Sitewise about incomplete queries and the constant Loading indicator while the user is editing the query is disruptive. 

This removes ALL calls to onRunQuery, making it necessary for the user to click on Run Query button in the panel if they want to send the query. 

Pros: 
- no unecessary unintentional calls to the backend
- in line with the Grafana-wide push to not run queries from plugin editors
- since calling onRunQuery() runs all queries in the panel, as does the Run Query button, this won't change the payload
Cons:
- user has to make an additional step

Another solution would be to add a `onFilterQuery` in the datasource, which will have to check every type of query in Sitewise and the necessary fields it needs.  We might still want that even if we merge this PR, but I think this PR also reduces the need to implement it. 

https://github.com/grafana/iot-sitewise-datasource/assets/16140639/b03e799a-6152-4876-86c9-e8b2def80a98

Fixes #

- Also enabled newFormStyling toggle for e2e tests, since we're removing it soon anyway. 

- Had to also replate the <Text/> component with a div since it's only available >grafana@10.2 and test were failing on earlier versions. This would also mean the component wouldn't render on those earlier versions once we remove the feature toggle.


kevinwcyu (commented):
I like the idea of not running the query on every change. Could we add the run query button like we have in the other data sources (e.g. cloudwatch). It might not be clear for the user that the refresh button in the query editor is meant for running the query.

idastambuk:
> I like the idea of not running the query on every change. Could we add the run query button like we have in the other data sources (e.g. cloudwatch). It might not be clear for the user that the refresh button in the query editor is meant for running the query.

I was considering this too, but it looks like we're moving away from having dedicated Run query buttons in specific editors and instead rely on using the refresh button or keyboard shortcut (there's more about it in this design doc: https://docs.google.com/document/d/1G62nVl3PNoot3mH9Ir74fLFzCKMftkd52dFATQsSxUo/edit#heading=h.cnw24jsf9gsq)

idastambuk:
@kevinwcyu added the header: 
Screenshot 2024-03-18 at 17 13 01

kevinwcyu (approved):
Cool, I think the e2e tests just need to click the Refresh panel button or the Run queries button to pass.

Issue #269, opened 2023-12-27
Feature request: Support for Asset Model component properties


mikejwil:
**Is your feature request related to a problem? Please describe.**
AWS IoT SiteWise now supports Asset model components to help industrial customers create reusable components to derive new asset models and express variations easily. For example, a wind turbine is made of many smaller components such as blades, rotors, and generators.  More info: [https://aws.amazon.com/about-aws/whats-new/2023/11/aws-iot-sitewise-supports-asset-model-components/](url)

Problem: Today, Grafana's iot-sitewise-datasource only allows for charting properties at the root of an asset, which means the properties within a component is not available for charting.

**Describe the solution you'd like**
Users should be able to select the properties grouped within a component.  To do so, the datasource should:

1. Instead of using DescribeAsset when retrieving the list of properties on an Asset, instead use [ListAssetProperties](https://docs.aws.amazon.com/iot-sitewise/latest/APIReference/API_ListAssetProperties.html) API with query parameter `filter` set to "ALL".
2. Use the `path` in the AssetPropertySummaries response object to showcase how properties are part of various components under the same asset (ex. `Motor1 /  AvgTorque` indicates the property is part of the Motor1 component versus `Torque` which a property defined at the root of the asset).

**Additional context**
Composite Models is used in the SiteWise API to define Components.  Therefore, the properties returned in step 1 will show the [assetCompositeModelId](https://docs.aws.amazon.com/iot-sitewise/latest/APIReference/API_AssetPropertySummary.html#iotsitewise-Type-AssetPropertySummary-assetCompositeModelId) it is part of, which is equivalent to a "component" in this context.

tracy-french:
@mikejwil Hi! I will take this issue!

mikejwil:
Great! Thanks @tracy-french.

seb-koch:
Hi @tracy-french , does the envisioned solution allow creating a variable that shows a list of all properties for a certain Assett? 

I am trying to understand if this is the right Issue here or if I should dig deeper or create a new issue for what we're trying to do.

**Our use case:**
A dashboard with two variables. One for the assett and one for the property. Our customers would select the assett (i.e. a Site) and a property (i.e. Humidity Reading Sensor 1).

tracy-french:
Hey @seb-koch! The plan is to utilize ListAssetProperties/ListAssetModelProperties to populate the drop-down list of asset properties. Currently, it is using DescribeAsset, which does not contain the component properties. 

[...]

PR #279, opened 2024-01-18
feat: support composite model properties

tracy-french:
**What this PR does / why we need it**: This change enables the visualization of component properties. Instead of utilizing the list of asset properties contained within DescribeAsset, ListAssetProperties is utilized. See https://github.com/grafana/iot-sitewise-datasource/issues/269.

### DIFF
--- a/src/components/query/PropertyQueryEditor.tsx
+++ b/src/components/query/PropertyQueryEditor.tsx
@@ -48,6 +48,7 @@ interface State {
   asset?: AssetInfo;
   property?: AssetPropertyInfo;
   assets: Array<SelectableValue<string>>;
+  assetProperties: Array<{ id: string; name: string }>;
   loading: boolean;
   openModal: boolean;
 }
@@ -57,6 +58,7 @@ const ALL_HIERARCHIES = '*';
 export class PropertyQueryEditor extends PureComponent<Props, State> {
   state: State = {
     assets: [],
+    assetProperties: [],
     loading: true,
     openModal: false,
   };
@@ -71,6 +73,8 @@ export class PropertyQueryEditor extends PureComponent<Props, State> {
     if (query?.assetIds?.length) {
       try {
         update.asset = await cache.getAssetInfo(query.assetIds![0]);
+        const ps = await cache.listAssetProperties(query.assetIds[0]);
+        update.assetProperties = ps.map(({ id, name }) => ({ id, name }));
       } catch (err) {
         console.warn('error reading asset info', err);
         update.property = undefined;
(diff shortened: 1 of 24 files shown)


src/components/query/PropertyQueryEditor.tsx
```diff
@@ -351,7 +355,7 @@ export class PropertyQueryEditor extends PureComponent<Props, State> {
 
     const isAssociatedAssets = isListAssociatedAssetsQuery(query);
     const showProp = !!(!isAssociatedAssets && (query.propertyId || query.assetIds));
-    const properties = showProp ? (asset ? asset.properties : []) : [];
+    const properties = showProp ? stateProps.map(({ id, name }) => ({ id, name, value: id, label: name })) : [];
```
sarahzinger, line 358:
```suggestion
    const displayProperties = showProp ? stateProps.map(({ id, name }) => ({ id, name, value: id, label: name })) : [];
```

again just trying to think of some other name to differentiate so we don't need stateprops (which is kind of confusing because react has both state and props)

src/components/query/PropertyQueryEditor.tsx
```diff
@@ -338,7 +342,7 @@ export class PropertyQueryEditor extends PureComponent<Props, State> {
 
   render() {
     const { query, datasource } = this.props;
-    const { loading, asset, assets } = this.state;
+    const { loading, assets, properties: stateProps } = this.state;
```
sarahzinger, line 345:
```suggestion
    const { loading, assets, properties } = this.state;
```

I think since props has an existing meaning in react, it's easy to be confused as to what this means.

sarahzinger (changes requested):
Hey there, thanks for the pr! Just want to make sure I understand. Is the idea here I should see all properties for an asset including properties of a child asset? If so I'm not sure I'm seeing that when I test it. 

Right now if I select CookieFactoryCameraAsset for example through the "Explore" folder, which I believe is a child asset of CookieFactoryHubAsset, I get around 10 properties. But then if I select it's parent CookieFactoryHubAsset, I still only get it's direct property of "hub".

Apologies if I may be misunderstanding the expected behavior here!

tracy-french:
@sarahzinger Thank you for reviewing! I'll follow-up on your requested changes shortly. :)

To answer your question, you should not see the asset properties of child assets. You should only see the properties of the selected asset(s), which is the same as the current behavior. The big difference is now the user can see component properties associated with the selected asset. Components are a new type of asset model which allows sharing of asset properties across asset models. A car asset model can have an engine component with a temperature asset property. The engine component can have a piston component with a pressure asset property. The engine component could be associated with a boat asset model as well. Both the boat and the car have temperature and pressure asset properties.

With the current implementation, we use DescribeAsset, which does not include the list of asset properties coming from associated components. Calling DescribeAsset would not show the temperature or pressure asset properties for both the boat and the car. ListAssetProperties does include these additional component asset properties.

sarahzinger:
Ahh ok thanks for clarifying that @tracy-french sorry for the misunderstanding! Your explanation makes a lot of sense thanks for walking it through! I was able to make a test asset with component models attached and see their properties! Looks great. When you're done making changes to this pr you can click the re-request review button on the top right hand side of this pr and it will ping me to take another look. Happy to merge once the little renames are done, unless @iwysiu has any other thoughts. Thanks!

Issue #161, opened 2023-02-06
failed to fetch query data: : Unknown Operation status code: 404, request id:


misiek303:
I am unable to query data anymore.  I get the below error:

```
failed to fetch query data: : Unknown Operation status code: 404, request id:
```
Grafana Logs:

```
logger=context userId=1 orgId=1 uname=admin t=2023-02-06T10:38:36.314059731-06:00 level=info msg="Request Completed" method=POST path=/api/ds/query status=400 remote_addr=192.168.50.132 time_ms=18 duration=18.094402ms size=137 referer="http://192.168.50.248:3000/d/YUUwUxA4k/txrx-simulator?orgId=1&refresh=5s" handler=/api/ds/query
```

**Environment**:
- Grafana version: 9.3.6
- Plugin version: 1.6.1
- OS Grafana is installed on: Ubuntu Server 20.10 LTE
- Data Processing Pack: 2.1.29

![Screenshot 2023-02-06 at 10 37 33 AM](https://user-images.githubusercontent.com/371344/217030313-662e70ec-8ee2-4ae1-b91b-5481bc11684c.png)

Connection to the Datasource works just fine

![Screenshot 2023-02-06 at 10 38 48 AM](https://user-images.githubusercontent.com/371344/217030615-ce1a059c-0dbc-41bc-becc-c98f0407126d.png)

I am also able to use AWS OpsHub and see all models, assets, and the latest data

<img width="1413" alt="Screenshot 2023-02-06 at 10 40 06 AM" src="https://user-images.githubusercontent.com/371344/217031355-8e508433-f17a-4d1f-8aea-805a83433ec1.png">

<img width="1429" alt="Screenshot 2023-02-06 at 10 40 22 AM" src="https://user-images.githubusercontent.com/371344/217031320-4f5373e3-d482-4b49-b59d-e19905f06d46.png">

Can you help ?

Vijay-Karthick:
I can confirm the same. With latest version, 'Get property value history' or any similar APIs were not displaying asset selection. In fact, the drop-down menu for asset selection and property selection were missing. I had to downgrade to 1.5.1 and everything started working as expected.

However, in latest version, list assets and list asset models were working fine.

katebrenner:
close as completed in https://github.com/grafana/iot-sitewise-datasource/pull/281

PR #281, opened 2024-01-24
fix: use non-batch APIs at the edge

tracy-french:
**What this PR does / why we need it**: IoT SiteWise BatchGet* APIs are not available at the edge. This change utilizes the non-batch variants of the APIs when at the edge.

**Which issue(s) this PR fixes**: https://github.com/grafana/iot-sitewise-datasource/issues/161


njvrzm (commented):
My many comments are mostly stylistic, but one or two more substantial things.

Overall I think the approach looks reasonable!

tracy-french:
@njvrzm Thank you for the early review!

I noticed most of your feedback is on the *_batch.go files, which is not new code. Those files are simply renamed to distinguish from the non-batch API variants. Do we need to address the issues you identified for the existing code in this PR?

njvrzm:
Indeed, my apologies, it was only towards the end of reviewing that I saw that most of the code is just moved 😅  I generally try to "leave it better" but that doesn't really apply here, and no, you don't need to address most of my feedback in this PR. I'll resolve some of my comments and maybe come back to them later myself.

idastambuk:
Hi @tracy-french, I think it makes sense to rename the files, but it could create problems down the line when we lose their entire source control history. Would it be possible for you to redo this PR in a way that preserves those files' history by renaming them with `git mv`?

tracy-french:
@idastambuk For sure! I'll make that change.

tracy-french:
@idastambuk I renamed the files with `git mv`, yet it didn't appear to make any difference from what I had before. Is it supposed to preserve the blame? Reading through https://stackoverflow.com/questions/1094269/whats-the-purpose-of-git-mv makes it seem like it's just a convenience function and git doesn't handle renames using it differently than a renames done outside of git?

njvrzm:
Hey @tracy-french, thanks for addressing the review feedback so far. I need to figure out how to trigger the drone build so we can make sure that check runs. Meanwhile I ran the tests locally and the tests in `TestPropertyValueAggregate` are failing - could you fix that please?

pkg/framer/property_values_for_time_range.go
```diff
@@ -2,20 +2,21 @@ package framer
 
 import (
```
idastambuk, line 3: I think the content of this got switched with pkg/framer/property_values_for_time_range_batch.go

pkg/framer/property_value_history.go
```diff
@@ -12,79 +12,39 @@ import (
...
-	} else {
-		frameName = *property.AssetName
-	}
-	frame := data.NewFrame(
-		frameName,
-		timeField,
-		valueField,
-		qualityField)
-
+	frame := data.NewFrame(*property.AssetName, timeField, valueField, qualityField)
```
idastambuk, line 35: Should we call `getFrameName()` here instead of *property.AssetName? I remember fixing this for "raw" queries, I assume it also applies for non-batch queries

idastambuk:
> @idastambuk I renamed the files with `git mv`, yet it didn't appear to make any difference from what I had before. Is it supposed to preserve the blame? Reading through https://stackoverflow.com/questions/1094269/whats-the-purpose-of-git-mv makes it seem like it's just a convenience function and git doesn't handle renames using it differently than a renames done outside of git?

Hi @tracy-french, yes, I noticed the same, I think it might be because the original files were renamed, but they were replaced with files with the same name (property_values_non_batch.go became the original property_values.go). I overlooked that when I suggested the solution, sorry about that!

tracy-french:
@idastambuk @njvrzm Feedback addressed! Thank you!

idastambuk:
Hi @tracy-french it looks like backend tests are failing, can you take a look? If you need help running them locally, let us know!

tracy-french:
@idastambuk Sorry about that! It looks like a forgot a file when creating the commit. Working on resolving now!

idastambuk (approved):

tracy-french:
@idastambuk @njvrzm Could we please merge this change in? Thank you! :)

PR #286, opened 2024-02-28
fix: asset property list render

tracy-french:
**What this PR does / why we need it**: https://github.com/grafana/iot-sitewise-datasource/pull/279 included a change which caused the asset property drop-down to not render the asset property names. This is due to a reference to the wrong list, which was not caught by TypeScript, as the `options` prop of `<Select />` does not have strong type safety (i.e., it only requires a list of type `T`).

https://github.com/grafana/iot-sitewise-datasource/pull/285 acts as an effective regression test to ensure asset property list is functional.


kevinwcyu (approved):

kevinwcyu:
Thanks for the fix @tracy-french, I'll have a look at #285 next.

PR #290, opened 2024-03-12
test: happy path playwright query tests

tracy-french:
**What this PR does / why we need it**: This change includes happy path playwright query tests for all seven queries exposed by the query editor. The work includes:
- Happy path test cases for each query using mocked data and making assertions on the rendering of expected form elements.
- Addition of `QueryEditor` page object model test utility used to reduce duplication within tests and make it easier to write and maintain tests.
- Addition of `queryEditor` test fixture to remove the need to instantiate the `QueryEditor` object model explicitly in every test.
- Addition of missing ARIA labels in query editor components used for testing selectors.

Next steps:
- Tests utilizing all of the available form fields (e.g., resolution, aggregation, time, format, etc.).
- Tests for failure cases.
- Improved request interception code (there is current significant duplication and code size).

### DIFF
--- a/src/components/query/ListAssetsQueryEditor.tsx
+++ b/src/components/query/ListAssetsQueryEditor.tsx
@@ -100,8 +100,9 @@ export class ListAssetsQueryEditor extends PureComponent<Props, State> {
     ) : (
       <>
         <div className="gf-form">
-          <InlineField label="Model ID" labelWidth={firstLabelWith} grow={true}>
+          <InlineField htmlFor="model" label="Model ID" labelWidth={firstLabelWith} grow={true}>
             <Select
+              inputId="model"
               isLoading={!models}
               options={modelIds}
               value={currentModel}
@@ -116,8 +117,9 @@ export class ListAssetsQueryEditor extends PureComponent<Props, State> {
           </InlineField>
         </div>
         <div className="gf-form">
-          <InlineField label="Filter" labelWidth={firstLabelWith} grow={true}>
+          <InlineField htmlFor="filter" label="Filter" labelWidth={firstLabelWith} grow={true}>
             <Select
+              inputId="filter"
               options={filters}
               value={filters.find((v) => v.value === query.filter) || filters[0]}
               onChange={this.onFilterChange}
--- a/tests/helpers.ts
+++ b/tests/helpers.ts
@@ -1,0 +1,17 @@
+import { test as base, expect } from '@grafana/plugin-e2e';
+import { QueryEditor } from './queryEditor.model';
+
+interface Fixtures {
+  queryEditor: QueryEditor;
+}
+
+const test = base.extend<Fixtures>({
+  /** Isolated `QueryEditorPage` instance. */
+  queryEditor: async ({ page }, use) => {
+    const queryEditor = new QueryEditor(page);
+
+    await use(queryEditor);
+  },
+});
+
+export { test, expect };
--- a/src/components/AggregatePicker.tsx
+++ b/src/components/AggregatePicker.tsx
@@ -84,7 +84,7 @@ export class AggregatePicker extends PureComponent<Props> {
     }
     return (
       <Select
-        id="aggregate-picker"
+        inputId="aggregate-picker"
         aria-label="Aggregate picker"
         value={select.current}
         className={className}
--- a/src/components/query/PropertyQueryEditor.tsx
+++ b/src/components/query/PropertyQueryEditor.tsx
@@ -259,7 +259,7 @@ export class PropertyQueryEditor extends PureComponent<Props, State> {
       </EditorFieldGroup>
     ) : (
       <div className="gf-form">
-        <InlineField label="Aggregate" labelWidth={firstLabelWith} grow={true}>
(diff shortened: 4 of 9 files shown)


tests/helpers.ts
```diff
@@ -0,0 +1,17 @@
...
+const test = base.extend<Fixtures>({
+  /** Isolated `QueryEditorPage` instance. */
+  queryEditor: async ({ page, panelEditPage }, use) => {
+    const queryEditor = new QueryEditorPage({ page, panelEditPage });
+
+    await use(queryEditor);
+  },
+});
+
+export { test, expect };
```
sarahzinger, line 17: hmmm.... Maybe this is a personal style preference and a lack of familiarity with playwright conventions but I am not sure this abstraction is easier to read/write tests with? It definitely dries up the code, but if a test does fail or you need to write your own test you now need to know to look in a helper file to find this fixture and then look at queryeditorpage and then there might be a worry that a small change in one could affect other tests?

I guess when it comes to tests I almost don't mind a bit of redundancy if it makes them easier to read. But I don't want to block for a personal preference. 

Idk what do you think @kevinwcyu in your research does this kind of best practice for playwright you think? I'm happy to go with whatever you all think makes the most sense.

sarahzinger (commented):
Cool stuff! Thanks for contributing again! I'll let Kevin take another look in case I'm missing anything. I've shared a few gut reactions on code-style but I'm not that experienced with playwright so I'm not confident they are useful opinions or not. Might be worth sticking with what's here and seeing how it goes? Will defer to @kevinwcyu for a final approval/thoughts.

src/components/query/ListAssetsQueryEditor.tsx
```diff
@@ -102,6 +102,7 @@ export class ListAssetsQueryEditor extends PureComponent<Props, State> {
         <div className="gf-form">
           <InlineField label="Model ID" labelWidth={firstLabelWith} grow={true}>
             <Select
+              aria-label="Model ID"
```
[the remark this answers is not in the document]
tracy-french: These ARIA labels are actually already used with the new style of components, but it seems we forgot to add them to both places in multiple instances. For example, see https://github.com/grafana/iot-sitewise-datasource/blob/main/src/components/query/ListAssetsQueryEditor.tsx#L74 for the model ID label.

tests/helpers.ts
tracy-french: Definitely, makes sense. I'm a fan of using fixtures, but this is indeed a very simple fixture and does not currently add much value beyond preventing you from having to instantiate the page object model.

As the test cases are expanded, something which may add a lot of value are test fixtures which reduce setup boilerplate when we're trying to test a particular state of the query editor multiple times. For instance, a `getPropertyValueQueryEditor` providing a query editor with the "Get property value" query pre-selected and exposing a page object model with the locators for form elements specific to only that query (i.e., no resolution/aggregation/etc.). Or even more valuable would be a fixture with an asset property (or multiple) already selected so that you can perform a series a tests without explicitly going through the flow of selecting an asset and then selecting a property in every test.

But those more valuable features don't need what I did in this PR now lol. If we want to wait until the value is clear, I am fine removing the fixture from this change. :)

chejimmy: I think this structure provides a good foundation for code standard/style 👍 We might not see the value on the current scale but doesn't hurt to keep it

chejimmy (approved):
LGTM! Lay down an excellent foundation and setting the code style for tests

tests/helpers.ts
kevinwcyu: I think it's useful to include this for interacting with the different select elements in the query editor, especially for the select elements that load options asynchronously. I'd be happy to keep it in since it does follow Playwright conventions.

tests/queryEditor.page.ts
```diff
@@ -0,0 +1,74 @@
+import { type Page, type Locator } from '@playwright/test';
+import { type PanelEditPage } from '@grafana/plugin-e2e';
+
+/**
+ * Query Editor page object model testing utility providing properties and
+ * methods for common locators and actions.
+ */
+export class QueryEditorPage {
```
kevinwcyu, line 8:
```suggestion
export class QueryEditor {
```

Minor nit: can this be named `QueryEditor` since it's a component within the `PanelEditPage`. It would be good to name this file `queryEditor.model.ts` along with this change.

kevinwcyu (commented):
Nice, the page object model for the query editor looks really good. Just a few minor suggestions.

src/components/query/ListAssetsQueryEditor.tsx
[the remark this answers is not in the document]
tracy-french: Updated all of the locators to use `page.getByLabel` and updated the JSX to use `inputId` where needed.

tests/queryEditor.page.ts
tracy-french: Done!

tests/helpers.ts
sarahzinger: Ok cool, let's give it a go, and see how it works out for us!

sarahzinger (approved):

PR #291, opened 2024-03-15
Add keywords

kevinwcyu:
Add keywords to improve search in the `/connections/add-new-connection` page of Grafana


iwysiu (approved):
Should we should include `sitewise` in the keywords?

sarahzinger (approved):

PR #302, opened 2024-03-29
Fix fetching asset properties

### LINKED ISSUE #301: Properties are not being fetched correctly for assets
**What happened**:
Resource Handler requests for the asset properties are failing with `failed to fetch query data: InvalidParameter: 1 validation error(s) found.- missing required field, ListAssetPropertiesInput.AssetId.`

I expect it to fetch the properties.

Screenshot 2024-03-29 at 2 40 11 PM

**Screenshots**

**Environment**:
- Grafana version:
- Plugin version:
- OS Grafana is installed on:
- User OS & Browser:
- Others:

iwysiu:
Fixes fetching asset properties. Instead of calling `util.GetAssetId` on the base query it should use the assetId on the wrapper query.

Fixes #301 

### DIFF
--- a/pkg/models/asset.go
+++ b/pkg/models/asset.go
@@ -22,8 +22,8 @@ type ListAssetsQuery struct {
 
 type ListAssociatedAssetsQuery struct {
 	BaseQuery
-	HierarchyId string `json:"hierarchyId,omitempty"`
-	LoadAllChildren bool `json:"loadAllChildren,omitempty"`
+	HierarchyId     string `json:"hierarchyId,omitempty"`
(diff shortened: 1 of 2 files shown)


pkg/models/asset.go
```diff
@@ -48,11 +48,7 @@ func GetListAssetPropertiesQuery(dq *backend.DataQuery) (*ListAssetPropertiesQue
 		return nil, err
 	}
 
-	// AssetId <--> AssetIds backward compatibility
-	query.MigrateAssetId()
```
sarahzinger, line 52: do we do this somewhere else now?

sarahzinger (approved):

PR #304, opened 2024-04-05
Migrate to CustomVariableSupport

### LINKED ISSUE #233: Migrate to CustomVariableSupport
**What happened**:
As described [here](https://github.com/grafana/iot-sitewise-datasource/pull/229), we need to investigate why this happens and reimplement custom variable support. 

> we need to use metricFindQuery for query variable support, otherwise the variable object resolves to {text: "Demo Wind Farm Asset", value "Demo Wind Farm Asset"}, which is not what we want, since when they are selected in dashboards, their IDs are sent in different types of queries and not their names.
> 
> The only time we would need the names is when the variables are being filtered by regex like https://github.com/grafana/iot-sitewise-datasource/issues/154. However, the solution for this in the original commit is just a consequence of the values being the same as names, and not a proper fix.
> I will reopen the original ticket, since we will need to reimplement custom variable support.

We should investigate why the values resolve to `{text: "Demo Wind Farm Asset", value "Demo Wind Farm Asset"}` and not to the assetId. For some reason the variable value in the redux state is equal to its text, for example:
Screenshot 2023-10-16 at 8 54 51 PM

The variable value should have value of the id. 

**Screenshots**

**Environment**:
- Grafana version:10.2.0
- Plugin version:1.10.0
- OS Grafana is installed on:
- User OS & Browser:
- Others:

iwysiu:
Migrates from the deprecated variable support to `CustomVariableSupport`. I created an [issue](https://github.com/grafana/iot-sitewise-datasource/issues/305) to actually design a new proper variable editor for this with resource handlers for the backend and an [issue](https://github.com/grafana/iot-sitewise-datasource/issues/306) to handle assetId and assetIds more consistently.

Fixes #233

### DIFF
--- a/src/DataSource.ts
+++ b/src/DataSource.ts
@@ -14,6 +14,7 @@ import { Observable } from 'rxjs';
 import { getRequestLooper, MultiRequestTracker } from 'requestLooper';
 import { appendMatchingFrames } from 'appendFrames';
 import { frameToMetricFindValues } from 'utils';
+import { SitewiseVariableSupport } from 'variables';
 
 export class DataSource extends DataSourceWithBackend<SitewiseQuery, SitewiseOptions> {
   // Easy access for QueryEditor
@@ -23,6 +24,7 @@ export class DataSource extends DataSourceWithBackend<SitewiseQuery, SitewiseOpt
   constructor(instanceSettings: DataSourceInstanceSettings<SitewiseOptions>) {
     super(instanceSettings);
     this.options = instanceSettings.jsonData;
+    this.variables = new SitewiseVariableSupport(this);
   }
 
   /**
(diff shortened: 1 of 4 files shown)


src/DataSource.ts
```diff
@@ -129,6 +131,7 @@ export class DataSource extends DataSourceWithBackend<SitewiseQuery, SitewiseOpt
       propertyAlias: templateSrv.replace(query.propertyAlias, scopedVars),
       region: templateSrv.replace(query.region || '', scopedVars),
       propertyId: templateSrv.replace(query.propertyId || '', scopedVars),
+      assetId: templateSrv.replace(query.assetId || '', scopedVars),
```
iwysiu, line 134: Without this, we can't use variables as asset ids

idastambuk (approved):
Looks good, Im getting properties! Too bad the variable interpolation is a mess rn ☹️

PR #313, opened 2024-05-24
chore: refactor paginator

chejimmy:
- simplify the SiteWise Queries pagination logic:
  - currently, the pagination logic under `src/requestLooper.ts` has a confusing call logic where it keeps creating new Observable to subscribe to in order to paginate exhaustively
  - scattered logic - `src/requestLooper.ts` attempted to generalize the logic by accepting 4 callbacks; it creates a fragmented logic between caller and requestLooper, hence, making the logic flow harder to follow
- this PR attempts to simplify the logic by:
  1. method `paginateQuery` paginate in a more tradition `do...while...` loop to handle asynchronous iteration; this is achievable after turning the `super.query(...)` into `Promise`; it is ok to do so since the SiteWise Query is not streaming and not necessary to apply the Observable pattern;
  1. streamlining the pagination logic under `src/SiteWiseQueryPaginator.ts` (see method `paginateQuery`); append frame data and getting next queries in the same method
- added unit testings for pagination - after the simplification, it allows the unit tests to be added easily

Fixes #

- `src/getNextQueries.ts` is copied from `src/DataSource.ts`


kevinwcyu (approved):
LGTM. 

Just need to fix the `yarn spellcheck` by fixing the typo `Pagainating` in `src/SiteWiseQueryPaginator.test.ts`. Also `ddced` in the `nextToken` is being picked up as an unknown word by the spellcheck. You can replace that substring in the token with some random characters to make it pass the spellcheck.

chejimmy:
> LGTM.
> 
> Just need to fix the `yarn spellcheck` by fixing the typo `Pagainating` in `src/SiteWiseQueryPaginator.test.ts`. Also `ddced` in the `nextToken` is being picked up as an unknown word by the spellcheck. You can replace that substring in the token with some random characters to make it pass the spellcheck.

🙏 TY for the review; updated for spellcheck 👍

cspell.config.json
```diff
@@ -62,6 +62,8 @@
     "errgroup",
     "ectx",
     "reqs",
-    "struct"
+    "struct",
+    "Pagainating",
+    "pagainating"
```
kevinwcyu, line 67: Could you actually remove these from the cspell exclusions, the correct spelling is `Paginating` without the extra "a" after the "g"

chejimmy: oh my bad, I didn't know how to spell
Lemme fix that 👍

chejimmy: All fixed 👍

kevinwcyu (approved):

PR #342, opened 2024-07-12
Remove ANY option for interpolated property quality

### LINKED ISSUE #326: Get interpolated property values quality "ANY"
**What happened**:
Updated the "Quality" setting of Get interpolated property values to "GOOD" then "ANY" and saw an error:
image

Data is successfully fetched for any quality.

1. Add a query for "Get interpolated property values"
2. Choose an asset and asset property (DOUBLE value type)
3. Change the "Quality" setting to "GOOD"
4. Change the "Quality" setting to "ANY"
5. See the error

**Screenshots**

https://github.com/grafana/iot-sitewise-datasource/assets/87385528/b1de01a8-e37e-4fa1-b58d-d647359fbc39

**Environment**:
- Grafana version: 9.4.0
- Plugin version: 1.12.1
- OS Grafana is installed on: MacOS
- User OS & Browser: Sonoma 14.5, Chrome
- Others: Amazon Managed Grafana environment

iwysiu:
When a "Get interpolated property values" query is used with `Quality` set to `ANY` (which is the default option) it errors because it's not a [valid value for the query](https://docs.aws.amazon.com/iot-sitewise/latest/APIReference/API_GetInterpolatedAssetPropertyValues.html#API_GetInterpolatedAssetPropertyValues_RequestSyntax). This PR removes it from the options for interpolated queries and automatically replaces `ANY` with `GOOD` on the backend (since it works on a first run and by using `GOOD` that's probably in line with users expect)

Fixes #326

### DIFF
--- a/src/components/query/QualityAndOrderRow.tsx
+++ b/src/components/query/QualityAndOrderRow.tsx
@@ -43,6 +43,8 @@ const qualities: Array<SelectableValue<SiteWiseQuality>> = [
   { value: SiteWiseQuality.UNCERTAIN, label: 'UNCERTAIN' },
 ];
 
+const interpolatedPropertyQualities: Array<SelectableValue<SiteWiseQuality>> = qualities.slice(1);
+
 const ordering: Array<SelectableValue<SiteWiseTimeOrder>> = [
   { value: SiteWiseTimeOrder.ASCENDING, label: 'ASCENDING' },
   { value: SiteWiseTimeOrder.DESCENDING, label: 'DESCENDING' },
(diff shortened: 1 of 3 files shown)


src/components/query/QualityAndOrderRow.tsx
```diff
@@ -104,14 +104,18 @@ export class QualityAndOrderRow extends PureComponent<Props> {
 
   render() {
     const { query } = this.props;
+    let currQualities = qualities;
+    if (query.queryType === QueryType.PropertyInterpolated) {
+      currQualities = qualities.slice(1);
+    }
```
kevinwcyu, line 110: nit:

add this to the top of the file underneath where `const qualities` is defined so it's only created once.

```
const interpolatedPropertyQualities: Array<SelectableValue<SiteWiseQuality>> = qualities.slice(1);
```

and change this here to

```suggestion
    const currQualities = query.queryType === QueryType.PropertyInterpolated ? interpolatedPropertyQualities : qualities;
```

kevinwcyu (approved):

src/components/query/QualityAndOrderRow.tsx
iwysiu: done

PR #352, opened 2024-09-17
fix: add check for nil for property value

ssjagad:

Currently, get property value and get property history error out when you query on a an asset with null or mixed value (null + value). We want to ensure make sure our experience is error free. 

Ensures that get property value and get property history to BAD or UNCERTAIN data is error free. 

https://github.com/user-attachments/assets/83e7b118-25bb-4247-b5aa-88352630357e

Fixes #


hwandersman (approved):

iwysiu (approved):

iwysiu:
There are ci errors that i have a pr up for: https://github.com/grafana/iot-sitewise-datasource/pull/353 and I'll merge it after those go in

PR #268, opened 2023-12-20
feat: implement support for sitewise sql

mitchlee-amzn:
**What this PR does / why we need it**:
This PR extends the capabilities of the datasource plugin to support the SiteWise ExecuteQuery API, which allows customers to leverage adhoc SQL queries against their assets and properties.

**Which issue(s) this PR fixes**:
https://github.com/grafana/iot-sitewise-datasource/issues/267

Fixes #

**Special notes for your reviewer**:

PR #364, opened 2024-10-22
Raw SQL Code Editor

jrgns:
This PR adds the ability to query IOT Sitewise data using the [query language](https://docs.aws.amazon.com/iot-sitewise/latest/userguide/sql.html)

This was previously attempted in https://github.com/grafana/iot-sitewise-datasource/pull/268 and allows users to use SQL to query data instead of the query builder. Using SQL allows for richer, customized queries that can combine data from multiple sources to be combined and presented in Grafana panels.

The UI when editing a panel. This shows a line graph of the returned data:

image

Notice the buttons in the header of the query, next to `Run queries` where the user can switch between the Builder (orinal) and Code (new).

The screenshot below shows the same information, but shows the raw data instead of the line graph:

image

And lastly the query and returned data in the Explorer view:

image

Delivery items

* [x] Simple query editor
* [x] Handler for SQL queries

See https://github.com/jadeit/iot-sitewise-datasource/pull/1 for 
* [x] Macro replacements ($__from, $__to, etc)
* [x] Code completion
* [x] Frontend Tests
* [x] Backend Tests


pkg/models/query.go
```diff
@@ -18,6 +18,7 @@ const (
 	QueryTypeDescribeAssetModel   = "DescribeAssetModel"
 	QueryTypeListAssetProperties  = "ListAssetProperties"
 	QueryTypeListTimeSeries       = "ListTimeSeries"
+	QueryTypeSQL                  = "SQL"
```
hwandersman, line 21: Let's use the API name `ExecuteQuery` for consistency

src/SitewiseQueryEditor.tsx
```diff
@@ -0,0 +1,91 @@
...
+            datasource={props.datasource}
+            query={query as SqlQuery}
+            onChange={onChangeInternal}
+            onRunQuery={onRunQueryInternal}
+          />
+        )}
+        {editorMode === QueryEditorMode.Builder && (
+          <VisualQueryBuilder
+            datasource={props.datasource}
+            query={query}
```
hwandersman, line 83: Swapping between modes will keep the query as a `SqlQuery`. Can consider storing the last selected query in the Builder mode to update the query when swapping between Code and Builder.

hwandersman:
High level comments:
1. My understanding is that the ExecuteQuery API is used to fetch SiteWise assets and asset metadata. This would be used as another option instead of the dropdowns / asset browser capability in the QueryEditor. Right now you have a toggle between the builder and code mode, but I think this should be contained within the builder mode as a way to find assets/properties to build a query to fetch data. This can be a feature of the asset browser so instead of the "Hierarchy" or "By Model" tabs there is a 3rd tab which is "By Query" or something like that.
2. Builder vs Code mode - Code mode as a standalone query misses all of the query options we have for our SiteWise queries (caching, L4E, etc.). Another reason the ExecuteQuery should be a sub-feature of asset selection in the asset browser.

PR #371, opened 2024-12-10
Chore: group dependabot updates on weekly interval

iwysiu:
- Groups updates according to the package ecosystem
- Sets update interval to be weekly (this will be Mondays by default)

I realized we never updated dependabot for this repo

Fixes: https://github.com/grafana/oss-plugin-partnerships/issues/927

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -20,8 +20,10 @@ updates:
     directory: '/'
     schedule:
       interval: 'weekly'
-    ignore:
-      - dependency-name: '@grafana/e2e*'
+      - dependency-name: 'react'
+        update-types: ['version-update:semver-major']
+      - dependency-name: 'react-dom'
+        update-types: ['version-update:semver-major']
     groups:
       all-node-dependencies:
         patterns:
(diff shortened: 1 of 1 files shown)


njvrzm (approved):
👍🏻👍🏻👍🏻

.github/dependabot.yml
```diff
@@ -1,13 +1,28 @@
...
+    groups:
+      all-github-action-dependencies:
+        patterns:
+          - '*'
+  - package-ecosystem: 'npm'
+    directory: '/'
+    schedule:
+      interval: 'weekly'
+    ignore:
+      - dependency-name: '@grafana/e2e*'
```
kevinwcyu, line 24:
```suggestion
      - dependency-name: 'react'
        update-types: ['version-update:semver-major']
      - dependency-name: 'react-dom'
        update-types: ['version-update:semver-major']
```

Might want to add this too, react just recently bumped its version to 19 and we'll probably want to stay on 18 at least until main grafana updates.

kevinwcyu (approved):

PR #390, opened 2025-01-15
feat: support asset selection by external id

### LINKED ISSUE #351: Support for SiteWise externalId on asset search

SiteWise released a feature to support the ability to provide an externalId to assets (https://docs.aws.amazon.com/iot-sitewise/latest/userguide/object-ids.html#external-ids).

Customers want to be able to select an asset based on externalId.

Currently in Grafana SiteWise, a user can select an asset based on an externalId, but the subsequent requests for timeseries data based on the asset selection fails as it attempts to utilize the externalId in the dataplane requests, a feature which is on the road map but not currently released.

The goal of this task is to make the changes necessary so that the dataplane requests for time-series data do not fail after selecting a asset based on externalId.

Below is the input where a user can enter in the externalId of an asset:
Image

It appears that currently grafana sitewise plugin utilizes the assetId the user enters into the input box as the asset id in the dataplane requests, causing the failure, since the dataplane does not support externalId yet. To resolve this task likely all that needs to occur is to utilize the assetId returned from the asset response, rather than the assetId entered into the text box.

---

Customer experience:

User can input the externalId directly into the asset input box. There will be 3 scenarios:

1. user explicitly add the `externalId: ` prefix, in which case the entered in string can be passed in directly to the request for the asset with no modification
2. user enters standard sitewise UUID as they would today (no modification, this behavior remains the same)
3. user enters in a string without the `externalId: ` prefix, and the string does not match the SiteWise UUID pattern (basic regex check client side is fine). In this case the front-end automatically appends the `externalId:` prefix before making the network call.

N/A

**Additional context**
N/A

mixu3:

Currently, the user is required to input only the asset id in uuid format in order to select the asset in a query. If any user tries to enter an [external id](https://docs.aws.amazon.com/iot-sitewise/latest/userguide/object-ids.html#external-ids), the query will not be able to find timeseries data since the dataplane does not support external ids yet.

This change allows users to select assets by external ids and retrieve the timeseries data.

https://github.com/user-attachments/assets/6a5b1e78-2c18-4a98-9e7b-c5a9871a2789

In addition, a tooltip for the asset field teaches users how to enter the asset id. Users can learn more about SiteWise external ids by clicking on the API docs button.

AssetTooltip

Fixes #351

**Testing**:
- Unit tests passing
- Manually testing these steps:
  - Typing in `externalId:...` retrieves the correct asset
  - Typing in the assetId in uuid format retrieves the correct asset (no change from previous behavior)
  - Typing in a string not in uuid format appends the `externalId:` prefix to retrieve the correct asset
  - Running queries on all three user inputs above loads the correct data
  - Successfully loads asset in query variable

### DIFF
--- a/src/components/query/PropertyQueryEditor.tsx
+++ b/src/components/query/PropertyQueryEditor.tsx
@@ -26,6 +26,8 @@ import { QueryOptions } from './QueryOptions';
 
 type Props = SitewiseQueryEditorProps<SitewiseQuery | AssetPropertyAggregatesQuery | ListAssociatedAssetsQuery>;
 
+const uuidRegex = /^[0-9a-f]{8}-[0-9a-f]{4}-[1-5][0-9a-f]{3}-[89abAB][0-9a-f]{3}-[0-9a-f]{12}$/i;
+
 const resolutions: Array<SelectableValue<SiteWiseResolution>> = [
   {
     value: SiteWiseResolution.Auto,
@@ -40,6 +42,7 @@ const resolutions: Array<SelectableValue<SiteWiseResolution>> = [
 ];
 
 interface State {
+  assetId?: string;
   asset?: AssetInfo;
   property?: AssetPropertyInfo;
   assets: Array<SelectableValue<string>>;
@@ -52,6 +55,7 @@ const ALL_HIERARCHIES = '*';
 
 export class PropertyQueryEditor extends PureComponent<Props, State> {
   state: State = {
+    assetId: this.props.query.assetIds && this.props.query.assetIds[0],
     assets: [],
     assetProperties: [],
     loading: true,
@@ -59,7 +63,7 @@ export class PropertyQueryEditor extends PureComponent<Props, State> {
   };
 
   async updateInfo() {
-    const { query, datasource } = this.props;
+    const { onChange, query, datasource } = this.props;
     const update: State = {
       loading: false,
     } as State;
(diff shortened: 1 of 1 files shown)


mixu3:
> Things to consider to ensure customers will understand this change:
> 
> 1. Label on the asset dropdown - should it still always be "Asset ID"?
> ...
> 2. Info icon on the **Asset** section with a description about how to select or enter an assetId.
> ...

After talking with UX, we agreed that we will keep the label as "Asset ID" but add an info icon to tell users how to enter the ID (UUID format or appending the prefix `externalId:`).

src/components/query/PropertyQueryEditor.tsx
```diff
@@ -168,7 +177,15 @@ export class PropertyQueryEditor extends PureComponent<Props, State> {
 
   onSetAssetId = (assetId?: string) => {
     const { onChange, query } = this.props;
-    onChange({ ...query, assetIds: assetId ? [assetId] : undefined });
+    if (!assetId) {
+      this.setState({ assetId: undefined });
+      onChange({ ...query, assetIds: undefined });
+    } else {
+      const assetIds =
+        uuidRegex.test(assetId) || assetId.startsWith('externalId:') ? [assetId] : [`externalId:${assetId}`];
```
idastambuk, line 185: I haven't tested this, but I assume that if the user is using a template variable, i.e $myAssetId, then this will append the `externalId` to it, which we don't want. The easiest solution to this would probably be to add a check for startsWith($)

chejimmy: ➕

mixu3: Good catch. Selecting the template variable from the dropdown did not have an issue but just typing the variable did. I updated and verified that it checks for '$' at the beginning of the asset id string

chejimmy (approved):
LGTM! TY for this new feature addition. I expect much customers appreciation :)

PR #394, opened 2025-01-17
fix: list associated assets query fields

### LINKED ISSUE #334: List Associated Assets supports hierarchyId, not property alias
**What happened**:
The List Associated Assets query has a field for `Property Alias` but this is not supported in the API.

This query should instead have a field for 'Hierarchy ID` as supported by the [SiteWise.ListAssociatedAssets](https://docs.aws.amazon.com/iot-sitewise/latest/APIReference/API_ListAssociatedAssets.html#API_ListAssociatedAssets_RequestSyntax) API.

The `Asset` field should be marked as required, with the optional field to specify a hierarchyId. We should list available hierarchies for the specified asset.

1. Add a List associated assets query in the query editor
2. Try to set the property alias
3. The asset field disappears and running the query leads to an error: `failed to fetch query data: InvalidParameter: 1 validation error(s) found. - missing required field, ListAssociatedAssetsInput.AssetId.`

**Screenshots**

**Environment**:
- Grafana version: 10.1.1
- Plugin version: 1.22.1
- OS Grafana is installed on: MacOS
- User OS & Browser: Sonoma 14.5, Chrome
- Others: Self Managed Grafana environment

mixu3:

Problem: The "List associated assets" query has a field for property alias but this is not supported in the API. When the property alias is set, the required asset field disappears.

Solution: Removing the property alias field and making sure the asset field is always visible for the "List associated assets" query (even if the property alias is set in another query) will solve this issue.

The query does use the hierarchy id, which is in the field called "Show." As this is confusing for users, I'm changing this field label to "Asset Hierarchy."

https://github.com/user-attachments/assets/6c3b9a28-ba75-4f9d-b884-c61c12496150

Fixes #334

**Testing**:
- Unit tests
- Manually testing:
  - Property alias field is never showing when query type is "List associated assets"
  - When property alias is defined elsewhere, the asset field does not disappear
  - Running this query works for hierarchy field set to "Parent", "All", and any hierarchies associated with the selected asset

### DIFF
--- a/src/components/query/PropertyQueryEditor.tsx
+++ b/src/components/query/PropertyQueryEditor.tsx
@@ -270,10 +270,10 @@ export class PropertyQueryEditor extends PureComponent<Props, State> {
       }
     }
     return (
-      <EditorField label="Show" htmlFor="show">
+      <EditorField label="Asset Hierarchy" htmlFor="assetHierarchy">
         <Select
-          id="show"
-          aria-label="Show"
+          id="assetHierarchy"
+          aria-label="Asset Hierarchy"
           isLoading={loading}
           options={hierarchies}
           value={current}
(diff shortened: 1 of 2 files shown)


hwandersman (approved):

chejimmy (approved):
LGTM! 🚀

Issue #398, opened 2025-01-21
Error message for List associated assets query variable


mixu3:
**What happened**:

When trying to create a query variable with query type "List associated assets," an error message pops up saying "Cannot read properties of undefined (reading 'length')." This is happening because as soon as the query type is selected, the query is being run. The user doesn't even have a chance to select an asset before the error message shows up.

**What you expected to happen**:

After selecting the query type "List associated assets," no error message should pop up right away. It should wait for the user to select an asset.

**How to reproduce it (as minimally and precisely as possible)**:

- Go to dashboard settings
- Choose tab "Variables"
- Choose "Add/New variable"
- Scroll down. Under query options, select a SiteWise data source
- Under query type, select "List associated assets"
- Error message should pop up on the top right corner

**Video recording**

https://github.com/user-attachments/assets/746d2a40-320d-43aa-b5b3-9fe1bdc10320

**Environment**:
- Grafana version: 10.4.5
- Plugin version: 1.25.1
- OS Grafana is installed on: MacOS
- User OS & Browser: Sequoia 15.2, Google Chrome
- Others: Self Managed Grafana environment

chejimmy:
does it block the customer from running the queries? Trying to assess the impact 🙏

PR #399, opened 2025-01-24
Bugfix: Filter out invalid queries in variable editor

idastambuk:

Template variable feature runs queries on change. This causes misleading errors for users, as [here](https://github.com/grafana/iot-sitewise-datasource/issues/398). This filters out queries in variable support for Sitewise so that only queries with defined required fields are run. 
(For queries in panels and explore, those have already been set up so they don't run automatically.)

Fixes https://github.com/grafana/iot-sitewise-datasource/issues/398

### DIFF
--- a/src/variables.ts
+++ b/src/variables.ts
@@ -1,4 +1,4 @@
-import { Observable } from 'rxjs';
+import { Observable, of } from 'rxjs';
 import { map } from 'rxjs/operators';
 import { assign } from 'lodash';
 import { QueryType, SitewiseQuery } from './types';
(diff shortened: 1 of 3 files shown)


src/variables.ts
```diff
@@ -17,15 +17,19 @@ export class SitewiseVariableSupport extends CustomVariableSupport<DataSource, S
...
-    assign(request.targets, [{ ...request.targets[0], refId: 'A' }]);
-    const response = this.datasource.query(request);
-    switch (request.targets[0].queryType) {
-      case QueryType.ListAssetModels:
-      case QueryType.ListAssets:
-      case QueryType.ListAssociatedAssets:
-        return this.parseOptions(response);
-      default:
-        return response;
+    if (this.isValidQuery(request.targets[0])) {
```
chejimmy, line 20: Nice! This should prevent unnecessary invocation of `query(...)`. 👍 
Curious if there's another place/interface to achieve this in Grafana?

chejimmy (approved):

src/variables.ts
chejimmy, line 20: Nice! This should prevent unnecessary invocation of `query(...)`.
Curious if there's another place/interface for handling query validations?

src/variables.ts
```diff
@@ -49,4 +53,20 @@ export class SitewiseVariableSupport extends CustomVariableSupport<DataSource, S
...
+      case QueryType.PropertyValue:
+      case QueryType.PropertyValueHistory:
+      case QueryType.PropertyInterpolated:
+      case QueryType.PropertyAggregate:
+        return Boolean((query.assetIds?.length || query.assetId) && query.propertyId);
+      case QueryType.ListAssetModels:
+      case QueryType.ListAssets:
+      case QueryType.ListAssociatedAssets:
+        return Boolean(query.assetIds?.length);
+      default:
```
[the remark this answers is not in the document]
idastambuk: Done!

src/variables.ts
idastambuk: There isn't one for queries in template variable editor. For normal queries there's `filterQuery` method [here](https://github.com/grafana/grafana/blob/763f0bac90d6296f034939627e9058bf2071b484/packages/grafana-data/src/types/datasource.ts#L276); however this quietly filters out the invalid query so it's normally only used for datasources where onChange triggers query request. In Sitewise, we've decided to go the way of not triggering a request automatically, so we don't really need the `filterQuery` yet.

PR #432, opened 2025-02-26
feat: Support nested variable UX for modelId

diehbria:
Utilizes new "nested template variable" UX in dropdown for modelId on the ListAssetQuery editor.

To help improve UX, and UX consistency.

Screenshot 2025-02-26 at 10 38 20 AM

N/A

Fixes #

- Reactivity bug

There is a minor reactivity bug for template variable values to flow down to the dropdown, where if a user changes the variables, it isn't immediately reflected in a dropdown.

This is a reactivity bug that is present already, a bit of a niche case that should be resolved but I think will require a larger refactor of dataflow to turn template variables to be passed in higher in the component tree that I want to tackle seperately.

- Cache refactor

I want to work towards uncoupling the caching module from options. To keep caching just about caching the resources, and have different area handle converting that data to options. This will be incremental work.

Also working on moving towards hook based helpers to remove the need for state management and memoization logic within the query editors.

### DIFF
--- a/src/sitewiseCache.ts
+++ b/src/sitewiseCache.ts
@@ -5,6 +5,7 @@ import { AssetModelSummary, AssetSummary, DescribeAssetResult } from './queryRes
 import { AssetInfo, AssetPropertyInfo } from './types';
 import { map } from 'rxjs/operators';
 import { getTemplateSrv } from '@grafana/runtime';
+import { useEffect, useState } from 'react';
 
 /**
  * Keep a different cache for each region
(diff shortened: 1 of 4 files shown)


src/components/query/PropertyQueryEditor.tsx
```diff
@@ -411,7 +411,7 @@ export class PropertyQueryEditor extends PureComponent<Props, State> {
           </EditorRow>
         )}
 
-        {showOptionsRow && (
```
diehbria, line 414: this is a mistake, will revert in revision

src/sitewiseCache.ts
```diff
@@ -117,20 +118,16 @@ export class SitewiseCache {
...
     }
 
-    return options.concat(
-      models.toArray().map((model) => ({
-        label: model.name,
-        value: model.id,
-        description: model.description,
-      }))
-    );
+    return models.toArray().map((model) => ({
```
idastambuk, line 126: (we could probably move this function getModelsOptions back to the component in the refactor you mentioned)

idastambuk:
Looks good, do we want to apply this to Asset dropdown while we're at it? (I think that's the only other one we support variables in, but I could be wrong)

diehbria:
> Looks good, do we want to apply this to Asset dropdown while we're at it? (I think that's the only other one we support variables in, but I could be wrong)

I choose not to touch the asset drop down because @hwandersman is working on changing the queries to support multiple asset / property pairs in a single query so I didn't want to conflict with his work

src/sitewiseCache.ts
diehbria: yes exactly the plan, trying not to turn this PR into the "big cache refactor" but it will come

idastambuk (approved):

PR #433, opened 2025-02-26
feat: support template var for interpolated resolution; make UX consi…

diehbria:
Add support for template variables and make UX consistent for the resolution field on the GetInterpolatedAssetPropertyValues query.

**BEFORE**
- resolution in query options, inconsistent with other query editors
- no support for template variables
Screenshot 2025-02-26 at 1 24 35 PM

**Example of aggregate history query panel, note the resolution option not hidden under query options**
Screenshot 2025-02-26 at 1 24 12 PM

**AFTER:**
After this change, template variable is supported, location is consistent, and as a bonus clicking the label correctly selected the input.
Screenshot 2025-02-26 at 2 11 08 PM

The PR does three things:
- Support for template variable for resolution query field for the "GetInterpolatedAssetPropertyValues" API request
- Move the resolution settings for interpolated values consistent with the aggregated value history API call, which has the resolution along side the property dropdown
- Utilize the updated nested UX for template variables for said drop down

N/A

- Working on refactoring the "Quality and order" component, which does so much more than quality and order. It has become a catch-all for many query options and requires some taming.
- wanted to remove the pattern of a blind spread with a `any` cast to instead pass down the props for the component to declare it's own children. The flexibility granted by the react component as a prop is not utilized anywhere and is just obfuscating the implementation with no benefit while ruining all type safety in the process.
- There needs to be a broader effort around the query type utilization, but I didn't want to tackle that here. I think we should move away from the global parent "SiteWiseQuery" type which contains a union across many queries, towards a discriminated union of each type.


src/components/query/visual-query-builder/InterpolatedResolutionSettings.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+export const InterpolatedResolutionSettings = ({
+  onChange,
+  query,
+}: {
+  query: AssetPropertyInterpolatedQuery;
+  onChange: (value: AssetPropertyInterpolatedQuery) => void;
+}) => {
+  const onResolutionChange = (sel: SelectableValue<string>) => {
+    onChange({ ...query, resolution: sel.value as SiteWiseResolution });
```
idastambuk, line 33: We're casting to SitewiseResolution here, but it's possible that the variable value is a string that isn't a valid resolution option. Do you think it would be worth it to add some kind of validation here (like an inline error) in that case (and otherwise we'd do a type guard)?

idastambuk (commented):
Just a few questions, and I think we'll need to modify the tests

src/components/query/visual-query-builder/InterpolatedResolutionSettings.tsx
diehbria: the challenge is it can be any sitewise resolution or a template variable, and there's no way i'm aware of to differentiate between a valid template variable expression vs just a random string

src/components/query/visual-query-builder/QueryOptions.tsx
```diff
@@ -55,10 +59,9 @@ export function QueryOptions({
               <Switch value={query.flattenL4e} onChange={onFlattenL4eChange} />
             </EditorField>
           )}
-          {shouldShowQualityAndOrderComponent(query.queryType) &&
-            (showProp || query.propertyAlias) &&
-            showQuality &&
-            qualityAndOrderComponent}
+          {shouldShowQualityAndOrderComponent(query.queryType) && (showProp || query.propertyAlias) && showQuality && (
+            <QualityAndOrderRow query={query} onChange={onChange} datasource={datasource} />
```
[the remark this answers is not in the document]
diehbria: no - though removing it from the props requires a larger refactor of typescript types. I'm actually completely deleteing this component in the next PR however: https://github.com/grafana/iot-sitewise-datasource/pull/434/files#diff-05da67e39f03638acb3347ddc116254bc05a1ede0f3c61a4f18dc54c4994be72

src/components/query/visual-query-builder/InterpolatedResolutionSettings.tsx
idastambuk: aren't these the only valid strings for resolution: https://github.com/grafana/iot-sitewise-datasource/blob/5231a1c691a30cbfbac1ba7acb59e6888d2eae58/src/types.ts#L42?

idastambuk:
> Just a few questions, and I think we'll need to modify the tests

Nvm it was compat check failing, my mistake

src/components/query/visual-query-builder/InterpolatedResolutionSettings.tsx
diehbria: yes, after the template gets resolved. With my change, `$resolution` for example would be a valid string input into the component

idastambuk: We could resolve it here, but if you think it's not worth the work, I'm happy to keep this

idastambuk (approved):

PR #434, opened 2025-02-27
Query editor refactor, support quality, time ordering, and response format in correct queries

### LINKED ISSUE #430: Grafana alerts not possible with iot-sitewise-datasource need custom query
**What happened**:
Attempting to setup Grafana alerts with a data point from iot-sitewise-datasource, but getting an error:

```
Something went wrong when evaluating this alert rule
[sse.readDataError] [A] got error: input data must be a wide series but got type long (input refid)
```

After asking the support team at Grafana, they state that this is not possible:

> I have checked internally, and it is not possible to set up an alert in this case since the query output includes string values. Grafana's alerting system only supports numerical values for conditions.
>  
> Additionally, since custom querying is not possible with AWS IoT SiteWise, we cannot modify or transform the data within alerts.
>  
> A potential workaround would be to load the IoT SiteWise data into a database, transform the data to convert string values into numerical indicators, and then query the database table to create alerts based on those transformed values.

Here is the screenshot of the response form GrafanaLabs:

Image

Need "custom querying" to be implemented as a feature.

I expect to be able to configure alerts on this data just like any other datasource.

Setup a alert in Grafana using AWS IoT SiteWise Datasource.

**Screenshots**

**Environment**:
- Grafana version: Grafana Cloud Pro
- Plugin version: 1.25.2
- OS Grafana is installed on: N/A
- User OS & Browser: Edge, Firefox, Chrome
- Others:

diehbria:
This is an overhaul of the property query editor.

There are fields that were not being exposed as options, which are valid options. And there were fields being exposed, that were not valid options. This includes quality, time ordering, and response format.

Additionally I have stripped away conditional rendering of various query options based on property information being filled out. There is no reason to prevent people from seeing options such as time ordering, prior to selecting a property.

Below is the state of the editors after this PR:

Screenshot 2025-02-27 at 10 33 42 AM

Screenshot 2025-02-27 at 10 34 10 AM
Screenshot 2025-02-27 at 10 33 59 AM
Screenshot 2025-02-27 at 10 33 53 AM
Screenshot 2025-02-27 at 10 33 48 AM
Query editors should expose the options where are configurable in respective queries, which currently they do not.
N/A

This PR also contains a larger refactor towards functional components, and breaking down settings into individual react components. Also an inversion of business logic towards leaf node components.

This PR is intended to be merged in after https://github.com/grafana/iot-sitewise-datasource/pull/433 is completed. Any feedback from prior PR will be rebased here.


diehbria:
need to adjust some of the server side code

PR #439, opened 2025-03-03
Add support for listing associated assets for multiple assets

### LINKED ISSUE #403: Allow the ability to list associated assets for multiple assets
Specifically in this file: https://github.com/grafana/iot-sitewise-datasource/blob/main/pkg/sitewise/api/list_associated_assets.go#L20, you are unable to list assets associated with a list of assets. This code only accepts a single assetID despite the grafana form taking in multiple assetIDs. This breaks the ability to create complex chains of queries. 

I would like it so when I put multiple assets into the list associated asset query, I can return the assets associated with each chosen asset. 

I've looked to create multiple queries for each parent asset, but then this does not scale well as I create many assets with many children.

**Additional context**

Image

Instead of only showing one sensor, it should show 3+ assets associated with the chosen assets.

kevinwcyu:

Adds support for handling multiple asset ids when using a "List associated assets" query.

Fixes #403

I created Power Plant model and added a Motor model as part of its hierarchy. The below is an example of a "List associated assets" query with multiple asset ids.

list-associated-assets-multiple-assets

The changes also handle the cases where an asset has multiple parents as shown in the diagram [here](https://docs.aws.amazon.com/iot-sitewise/latest/userguide/asset-hierarchies.html).

Some notes on the changes:
- `pkg/server/handlers_test.go` added `assetIds` to the request since the updated code in `pkg/sitewise/api/list_associated_assets.go` loops through all of the asset ids to handle multiple assets, without this change the test fails since the expected call to the sitewise client isn't made if there are no assetIds in the request.
- Changes in `pkg/testdata/*` files: removed two of the assets from the response in `pkg/testdata/list-associated-assets.json` into `pkg/testdata/list-associated-assets.json` so that each file is associated with one asset id in the tests.
- I suggest viewing the diff with whitespace changes hidden.

### DIFF
--- a/pkg/sitewise/api/list_associated_assets.go
+++ b/pkg/sitewise/api/list_associated_assets.go
@@ -9,7 +9,6 @@ import (
 	"github.com/grafana/iot-sitewise-datasource/pkg/framer"
 	"github.com/grafana/iot-sitewise-datasource/pkg/models"
 	"github.com/grafana/iot-sitewise-datasource/pkg/sitewise/client"
-	"github.com/grafana/iot-sitewise-datasource/pkg/util"
 )
 
 func ListAssociatedAssets(ctx context.Context, client client.SitewiseClient, query models.ListAssociatedAssetsQuery) (*framer.AssociatedAssets, error) {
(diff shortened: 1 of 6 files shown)


pkg/sitewise/api/list_associated_assets.go
```diff
@@ -9,45 +9,91 @@ import (
...
-		assetId            *string = util.GetAssetId(query.BaseQuery)
 		results            []*iotsitewise.AssociatedAssetsSummary
 	)
 
-	// Recursively load children
-	if query.LoadAllChildren {
-		asset, err := client.DescribeAsset(&iotsitewise.DescribeAssetInput{
-			AssetId: assetId,
-		})
+	seenAssetIds := make(map[string]bool)
```
iwysiu, line 22: Neat Go fact: you could also use `map[string]struct{}` to make a set. I think it's technically slightly better for memory (though on this scale its more of a fun fact that real savings)

pkg/sitewise/api/list_associated_assets.go
```diff
@@ -9,45 +9,91 @@ import (
...
+
+			for _, h := range asset.AssetHierarchies {
+				var nextToken *string = nil
+
+				for {
+					resp, err := client.ListAssociatedAssetsWithContext(ctx, &iotsitewise.ListAssociatedAssetsInput{
+						AssetId:     assetIdPtr,
+						HierarchyId: h.Id,
+						MaxResults:  MaxSitewiseResults,
+						NextToken:   nextToken,
```
iwysiu, line 45: I noticed that the original code use "" for next token for the first case and doesn't return the token and for the other case used `getNextToken(query.BaseQuery)` and did return it. Was the second one a bug?

iwysiu (approved):
I tried it and it worked!

pkg/sitewise/api/list_associated_assets.go
kevinwcyu: The original code would handle fetching all of the associated assets so it didn't have to return a next token. The other case didn't have that handling before.

PR #440, opened 2025-03-04
fix: remove unsupported regions

tracy-french:

The datasource currently allows the user to select a large number of regions which do not support the AWS IoT SiteWise service.

Attempting to run a query in a region not supported by IoT SiteWise, such as the first region listed in the region drop-down, ap-east-1, leads to the following error:

```sh
failed to fetch query data: RequestError: send request failed caused by: Get "https://api.iotsitewise.ap-east-1.amazonaws.com/asset-models?maxResults=250": dial tcp: lookup api.iotsitewise.ap-east-1.amazonaws.com on 10.0.0.2:53: no such host
```

This PR removes the unsupported regions from the drop-down.

Additionally, as an improvement, the change enforces region type-safety via the `Region` type, a union of the string literal region names.

### DIFF
--- a/src/regions.ts
+++ b/src/regions.ts
@@ -1,35 +1,36 @@
-import { SelectableValue } from '@grafana/data';
+import { type SelectableValue } from '@grafana/data';
 
-export const standardRegions = [
-  'ap-east-1',
-  'ap-northeast-1',
-  'ap-northeast-2',
-  'ap-northeast-3',
+// see https://docs.aws.amazon.com/general/latest/gr/iot-sitewise.html#iot-sitewise_region-sdk
+// order based on order in documentation link
+export const supportedRegions = [
+  'us-east-2',
+  'us-east-1',
+  'us-west-2',
   'ap-south-1',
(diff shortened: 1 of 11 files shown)


tracy-french:
Tested locally. Not sure why the actions are failing. Looks like it could be a permissions issue? `Error: Error message: Unable to get ACTIONS_ID_TOKEN_REQUEST_URL env variable` in the output.

src/regions.ts
```diff
@@ -1,35 +1,35 @@
...
+  'us-east-2',
+  'us-east-1',
+  'us-west-2',
   'ap-south-1',
+  'ap-northeast-2',
   'ap-southeast-1',
   'ap-southeast-2',
+  'ap-northeast-1',
   'ca-central-1',
-  'cn-north-1',
```
diehbria, line 12: according to https://www.aws-services.info/iotsitewise.html cn-north-1 is supported

tracy-french: Fixed!

diehbria (approved):

PR #449, opened 2025-03-12
fix: client cache switch should control the correct query

### LINKED ISSUE #450: Client cache switch should control the correct query
**What happened**:

Client cache switch toggles the first query target only

Client cache switch toggles the corresponding query target

1. Create a SiteWise query ("A")
2. Duplicate the query ("B")
3. Toggle Client cache of the duplicated query ("B")
4. Toggle Client cache of the original query ("A")
5. Notice only the original query Client cache state is changing

**Screenshots**

https://github.com/user-attachments/assets/53f64f64-bd47-4270-9782-3e7fff88a50d

For some reason, when there are more than 1 `<Switch>`  with an `id`, only the `onChange` of the first `<Switch>` is triggered.

https://github.com/user-attachments/assets/756ad291-6097-4d88-b270-a1b88c02a502

**Environment**:
- Grafana version:
- Plugin version:
- OS Grafana is installed on:
- User OS & Browser:
- Others:

chejimmy:

This PR removed the `id` attribute from `<Switch>` of client cache to avoid only the first `<Switch>`'s `onChange` is triggered. This fixes the issue #450  

Fixes #450

### DIFF
--- a/src/components/query/visual-query-builder/ClientCacheRow.tsx
+++ b/src/components/query/visual-query-builder/ClientCacheRow.tsx
@@ -5,18 +5,21 @@ import { EditorField, EditorFieldGroup, EditorRow } from '@grafana/plugin-ui';
 interface Props {
   clientCache?: boolean;
   onClientCacheChange: (evt: React.SyntheticEvent<HTMLInputElement>) => void;
+  queryRefId: string;
 }
 
-export const ClientCacheRow = ({ clientCache, onClientCacheChange }: Props) => {
+export const ClientCacheRow = ({ clientCache, onClientCacheChange, queryRefId }: Props) => {
+  const cacheSwitchId = `client-cache-switch-${queryRefId}`;
+
   return (
     <EditorRow>
       <EditorFieldGroup>
         <EditorField
           label="Client cache"
-          htmlFor="clientCache"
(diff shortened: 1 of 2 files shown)


src/components/query/visual-query-builder/ClientCacheRow.tsx
```diff
@@ -16,7 +16,7 @@ export const ClientCacheRow = ({ clientCache, onClientCacheChange }: Props) => {
           htmlFor="clientCache"
           tooltip="Enable to cache results in the browser that are older than 15 minutes. Note: Dashboard variable query result will not update when client cache is enabled."
         >
-          <Switch id="clientCache" value={clientCache} onChange={onClientCacheChange} />
+          <Switch value={clientCache} onChange={onClientCacheChange} />
```
kevinwcyu, line 19: The `ClientCacheRow` could have a prop that takes the `refId` of the query. You can generate the `id` for the Switch like this

```ts
const id = `clientCache-${refId}`
```

And update the `htmlFor` on the `EditorField` to use that `id` and use it as the `id` for the `Switch` component.

This would still allow clicking on the label to toggle the switch.

kevinwcyu (changes requested):

PR #457, opened 2025-03-28
chore: grafana instance version support warnning

chejimmy:

We have incidents that customers using Grafana instances with versions prior to v10 are unable to see plugin versions beyond 1.25.2. To resolve this, we recommend that these customers upgrade their Grafana instances to version 10 or higher. This will ensure they can access the latest plugin versions and functionality.
To notify customers about the need to upgrade, we suggest adding a prominent message on the plugin page informing users that versions beyond 1.25.2 require a Grafana instance of v10 or later. This will help raise awareness and guide customers to the necessary upgrade to continue benefiting from the Grafana SiteWise plugin.

Render of the readme with the warning message.

Screenshot 2025-03-28 at 12 37 48 PM

Fixes #

### DIFF
--- a/src/README.md
+++ b/src/README.md
@@ -1,3 +1,7 @@
+> ***Warning***
+>
+> We recommend customers to use Grafana version 10.4.0 or higher for the AWS IoT SiteWise Datasource. Grafana instances with versions prior to 10.4.0 are unable to use AWS IoT SiteWise data source versions beyond 1.25.2.
+
 # AWS IoT SiteWise Datasource
 
 This datasource supports reading data from [AWS IoT SiteWise](https://aws.amazon.com/iot-sitewise/) and showing it in a Grafana dashboard.
--- a/src/plugin.json
+++ b/src/plugin.json
@@ -37,7 +37,7 @@
     "updated": "%TODAY%"
   },
   "dependencies": {
-    "grafanaDependency": ">=10.3.0",
+    "grafanaDependency": ">=10.4.0",
     "plugins": []
   }
 }


idastambuk:
Hi @chejimmy, I just noticed there's a bit of a discrepancy and went to double check. We use a component in the query editor from Grafana 10.4. [This readme](https://github.com/grafana/iot-sitewise-datasource/blob/main/README.md) says it's >=10.4. and that's indeed the minimum supported version. Can we update this to reflect that? 
The discrepancy is also in the plugin.json file where minimum Grafana version says to be 10.3., can we update that to be 10.4. as well? Just wanna make sure we have the correct info everywhere. 
Thanks a lot!

chejimmy:
> Hi @chejimmy, I just noticed there's a bit of a discrepancy and went to double check. We use a component in the query editor from Grafana 10.4. [This readme](https://github.com/grafana/iot-sitewise-datasource/blob/main/README.md) says it's >=10.4. and that's indeed the minimum supported version. Can we update this to reflect that? The discrepancy is also in the plugin.json file where minimum Grafana version says to be 10.3., can we update that to be 10.4. as well? Just wanna make sure we have the correct info everywhere. Thanks a lot!

Updated (see [diff](https://github.com/grafana/iot-sitewise-datasource/compare/49c376a443ba7d70e821a3c807da123a3c5a3852..a57a555caf5ffb72aea846b9e88960b89

This should fix the new version from now on;

However, previous versions (`2.1.0` and `2.0.0`) would not be changed and would maintain the same incorrect `grafanaDependency`; Want to learn if there's ways to patch releases? 🤔 (Might not be something we need to do but want to learn about the process)

Screenshot 2025-03-31 at 11 01 27 AM
463d8e5))

PR #402, opened 2025-01-31
Migrate to aws-sdk-go-v2

njvrzm:
**What this PR does / why we need it**:
This updates the datasource to use aws-sdk-go-v2, since the older version is in maintenance mode and will be end-of-lifed soon.

**Special notes for your reviewer**:
Would really appreciate any help you can give with testing this! I don't have the ability to exercise the datasource thoroughly.

PR #476, opened 2025-05-12
add support for edge

kevinwcyu:
Made a couple of changes to fix the edge region support when updating aws-sdk-go to v2.


pkg/framer/property_value_history.go
```diff
@@ -44,7 +44,7 @@ func (p AssetPropertyValueHistory) Frames(ctx context.Context, resources resourc
 		if v.Value != nil && getPropertyVariantValue(v.Value) != nil {
 			timeField.Set(i, getTime(v.Timestamp))
 			valueField.Set(i, getPropertyVariantValue(v.Value))
-			qualityField.Set(i, v.Quality)
+			qualityField.Set(i, string(v.Quality))
```
kevinwcyu, line 47: The edge region doesn't use the batch api. This change follows what's done in the main v2 upgrade PR https://github.com/grafana/iot-sitewise-datasource/pull/402/files#diff-fbd40360883c5e3ae50002127791ee48cd59f80b5f4d7292f12454446a4a7308R116

pkg/sitewise/client/client.go
```diff
@@ -299,8 +299,5 @@ func GetHTTPClient(settings models.AWSSiteWiseDataSourceSetting) (*http.Client,
 		},
 	}
 
-	// TODO: figure out how to replace this with smithy's version
-	// https://pkg.go.dev/github.com/aws/smithy-go/transport/http#DisableEndpointHostPrefix
-	//swcfg = swcfg.WithDisableEndpointHostPrefix(true)
```
kevinwcyu, line 304: This is covered by the change in pkg/sitewise/datasource.go

pkg/sitewise/datasource.go
```diff
@@ -30,6 +33,21 @@ type Datasource struct {
...
+
+func (m *disableHostPrefixMiddleware) HandleInitialize(
+	ctx context.Context, in middleware.InitializeInput, next middleware.InitializeHandler,
+) (
+	out middleware.InitializeOutput, metadata middleware.Metadata, err error,
+) {
+	ctx = smithyhttp.SetHostnameImmutable(ctx, true)
+	return next.HandleInitialize(ctx, in)
+}
+
```
kevinwcyu, line 50: This is to prevent the endpoint from automatically getting `api` prepended to it. It's used below only if the region is set to Edge.

E.g. when you configure the endpoint in the config editor to be something like `https://localhost:8443`, having this middleware prevents the client from making a call to `https://api.localhost:8443` instead.

pkg/sitewise/datasource.go
```diff
@@ -123,7 +148,7 @@ func (ds *Datasource) invoke(ctx context.Context, _ *backend.QueryDataRequest, b
 
 func (ds *Datasource) HealthCheck(ctx context.Context, req *backend.CheckHealthRequest) error {
 
-	sw, err := ds.getClient(ctx, "") // Default region
+	sw, err := ds.getClient(ctx, ds.cfg.Region)
```
kevinwcyu, line 151: Needed this to fix the "Save and Test" button in the config editor. The client needed to be configured with the Edge region instead of an empty region, otherwise I was getting a 403 error.

pkg/sitewise/datasource.go
```diff
@@ -134,7 +159,7 @@ func (ds *Datasource) HealthCheck(ctx context.Context, req *backend.CheckHealthR
 }
 
 func (ds *Datasource) HandleInterpolatedPropertyValueQuery(ctx context.Context, _ *backend.QueryDataRequest, query *models.AssetPropertyValueQuery) (data.Frames, error) {
-	sw, err := ds.GetClient(ctx, query.BaseQuery.AwsRegion)
+	sw, err := ds.getClient(ctx, query.BaseQuery.AwsRegion)
```
kevinwcyu, line 162: Fixing this https://github.com/grafana/iot-sitewise-datasource/pull/402#discussion_r2021448316

kevinwcyu (commented):
fyi @njvrzm, these were the fixes i needed to get the v2 branch to work with sitewise edge.

PR #539, opened 2025-07-30
add downstream error source in failed reques

ktw4071:
ErrorBudgetBurnRateTooHigh warnings in operation channel. These errors should be downstream errors, not plugin errors
https://raintank-corp.slack.com/archives/C077F654GCS/p1753890062885989
N/A

Fixes #

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 All notable changes to this project will be documented in this file.
 
+## 2.3.1
+
+- Chore: Add donwstream error for failed requests and unmarshalling
+
 ## 2.3.0
 
 - fix interpolated property value next token in [#506](https://github.com/grafana/iot-sitewise-datasource/pull/506)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@ All notable changes to this project will be documented in this file.
 
 ## 2.3.1
 
-- Chore: Add donwstream error for failed requests and unmarshalling
+- Chore: Add downstream error for failed requests and unmarshalling
 
 ## 2.3.0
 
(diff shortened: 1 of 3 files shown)


pkg/server/server.go
```diff
@@ -39,12 +39,14 @@ type QueryHandlerFunc func(context.Context, *backend.QueryDataRequest, backend.D
 func DataResponseErrorUnmarshal(err error) backend.DataResponse {
 	return backend.DataResponse{
 		Error: errors.Wrap(err, "failed to unmarshal JSON request into query"),
+		ErrorSource: backend.ErrorSourceDownstream,
```
iwysiu, line 42: I think it makes sense to leave unmarshalling errors as plugin errors: we should unmarshal correctly even if the request fails

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 All notable changes to this project will be documented in this file.
 
+## 2.3.1
```
iwysiu, line 5: can you add the other changes? You can use the release drafter tool to get them: https://github.com/grafana/iot-sitewise-datasource/blob/main/CONTRIBUTING.md#build-a-release

pkg/server/server.go
ktw4071: As I recall, we define marshall/unmarshall error as downstream error in our repo. But I can take off errorsource for unmarshalling.

CHANGELOG.md
ktw4071: I've never used this tool before. What should go in Release title, Tag, description?

image

iwysiu: Yeah, i tried the tool and it seems to be broken, I did i manually and pushed up a commit for it

iwysiu (approved):

Issue #551, opened 2025-08-08
Missing region and DescribeTimeSeries API call 404


kevinwcyu:
From version 2.3.0 of the data source, there's been some errors seen in Grafana Cloud where the region in the configuration is missing as well as the DescribeTimeSeries API call failing with a 404 error.

Example missing region error
```
failed to fetch query data: operation error IoTSiteWise: BatchGetAssetPropertyAggregates, failed to resolve service endpoint, endpoint rule error, Invalid Configuration: Missing Region
```

Example DescribeTimeSeries 404 error
```
failed to fetch query data: operation error IoTSiteWise: DescribeTimeSeries, https response error StatusCode: 404
```

PR #556, opened 2025-08-13
QueryData: Set default region if empty, prepare 2.3.2

idastambuk:
Fixes https://github.com/grafana/iot-sitewise-datasource/issues/551
Also set isClearable to false, since it shouldn't be possible for users to clear the region, since Default is one of the options

Screenshot 2025-08-13 at 10 56 32

Screenshot 2025-08-13 at 10 18 47

### DIFF
--- a/pkg/server/handlers.go
+++ b/pkg/server/handlers.go
@@ -13,10 +13,10 @@ import (
 	"github.com/grafana/iot-sitewise-datasource/pkg/models"
 )
 
-func getErrorResponse(err error, message string) *backend.QueryDataResponse {
+func getErrorResponse(err error) *backend.QueryDataResponse {
 	resp := backend.Responses{
 		"A": backend.DataResponse{
-			Error: errors.New(message + ": " + err.Error()),
+			Error: err,
 		},
 	}
 	return &backend.QueryDataResponse{Responses: resp}
@@ -34,10 +34,12 @@ func processQueries(ctx context.Context, req *backend.QueryDataRequest, handler 
 		req.PluginContext.DataSourceInstanceSettings.JSONData != nil {
 
 		if err := json.Unmarshal(req.PluginContext.DataSourceInstanceSettings.JSONData, &datasourceJsonData); err != nil {
-			return getErrorResponse(err, "failed to unmarshal datasource JSON data")
+			return getErrorResponse(errors.New("failed to unmarshal datasource JSON data: " + err.Error()))
 		}
 		if datasourceJsonData.DefaultRegion != "" {
 			defaultRegion = datasourceJsonData.DefaultRegion
+		} else {
+			return getErrorResponse(errors.New("Invalid query: default region is not set in datasource settings"))
 		}
 	}
 
@@ -45,7 +47,7 @@ func processQueries(ctx context.Context, req *backend.QueryDataRequest, handler 
 		var queryJSON map[string]interface{}
 
 		if err := json.Unmarshal(q.JSON, &queryJSON); err != nil {
-			return getErrorResponse(err, "failed to unmarshal query JSON")
+			return getErrorResponse(errors.New("failed to unmarshal query JSON: " + err.Error()))
 		}
 		region, ok := queryJSON["region"].(string)
 		if !ok || region == "" || region == "default" {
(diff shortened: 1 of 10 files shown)


pkg/server/handlers.go
```diff
@@ -11,10 +13,51 @@ import (
...
+	if req.PluginContext.DataSourceInstanceSettings != nil &&
+		req.PluginContext.DataSourceInstanceSettings.JSONData != nil {
+
+		if err := json.Unmarshal(req.PluginContext.DataSourceInstanceSettings.JSONData, &datasourceJsonData); err != nil {
+			return getErrorResponse(err, "failed to unmarshal datasource JSON data")
+		}
+		if datasourceJsonData.DefaultRegion != "" {
+			defaultRegion = datasourceJsonData.DefaultRegion
+		}
+	}
```
beejeebus, line 42: can we error out early here if `defaultRegion` is still the empty string?

beejeebus (approved):
left a question about handling empty default region, but otherwise LGTM.

pkg/server/handlers.go
```diff
@@ -11,10 +13,51 @@ import (
...
+
+	defaultRegion := ""
+	if req.PluginContext.DataSourceInstanceSettings != nil &&
+		req.PluginContext.DataSourceInstanceSettings.JSONData != nil {
+
+		if err := json.Unmarshal(req.PluginContext.DataSourceInstanceSettings.JSONData, &datasourceJsonData); err != nil {
+			return getErrorResponse(err, "failed to unmarshal datasource JSON data")
+		}
+		if datasourceJsonData.DefaultRegion != "" {
```
iwysiu, line 39: Double checking: is this one of the datasources where default region can be "default"? I know it can in cloudwatch

pkg/server/handlers.go
```diff
@@ -11,10 +13,51 @@ import (
...
+			defaultRegion = datasourceJsonData.DefaultRegion
+		}
+	}
+
+	for _, q := range req.Queries {
+		var queryJSON map[string]interface{}
+
+		if err := json.Unmarshal(q.JSON, &queryJSON); err != nil {
```
iwysiu, line 47: nit: I don't really love having to unmarshal and re-marshal the query (doing this can be a little expensive and we presumably need to unmarshal it again) but I do get doing the suboptimal thing to get this through)

iwysiu (approved):

pkg/server/handlers.go
```diff
@@ -11,10 +13,51 @@ import (
...
+
+	defaultRegion := ""
+	if req.PluginContext.DataSourceInstanceSettings != nil &&
+		req.PluginContext.DataSourceInstanceSettings.JSONData != nil {
+
+		if err := json.Unmarshal(req.PluginContext.DataSourceInstanceSettings.JSONData, &datasourceJsonData); err != nil {
+			return getErrorResponse(err, "failed to unmarshal datasource JSON data")
+		}
+		if datasourceJsonData.DefaultRegion != "" {
+			defaultRegion = datasourceJsonData.DefaultRegion
```
kevinwcyu, line 40: Can this be done here instead https://github.com/grafana/iot-sitewise-datasource/blob/ee4a1f1416fdf92367d4b6550e23512c5d8a8cf4/pkg/sitewise/datasource.go#L115 with something like

```
	if region == "" {
		region = ds.cfg.Region
	}
```

kevinwcyu: i think the `ds.cfg.Region` should already have the defaultRegion set there by the time we reach that code because it's loaded when the settings are loaded https://github.com/grafana/iot-sitewise-datasource/blob/ee4a1f1416fdf92367d4b6550e23512c5d8a8cf4/pkg/models/setting.go#L31-L33

idastambuk: Ok I think it should be possible, I thought region was necessary for the query but it isn't. Let me try to rewrite this

pkg/server/handlers.go
idastambuk: Managed to avoid it by applying Kevin's suggestion https://github.com/grafana/iot-sitewise-datasource/pull/556#discussion_r2273593932

PR #561, opened 2025-08-19
Added support for selecting multiple asset properties with wildcard.

### LINKED ISSUE #283: Get multiple properties from an asset, ideally with regex or wildcards
We're trying to setup dashboards for a customer where we are (planning to) roll out SiteWise for 200 locations. Right now we have to know the **exact property name and create visualizations for specific properties**. 

Typically, in most Grafana data sources, you can use regular expressions or general wildcards to select **multiple data properties and visualize these**, i.e. by showing all data points for anything that matches ```Ping*```.

The SiteWise plugin seems to have no support for this or a similar functionality that would do the job. I hope its just me and I don't get the usage right 😄 , and there is a way to create a panel that shows all the time series for a set of properties (i.e. all current measurements across all engines).

Support for regular expressions in the **Property** field and/or in the **Property Alias** field.
Screenshot 2024-02-07 at 17 46 04
Screenshot 2024-02-07 at 17 45 54

One could expose the ```ListAssettProperties``` [endpoint](https://docs.aws.amazon.com/iot-sitewise/latest/APIReference/API_ListAssetProperties.html) as a new query type, and we could use this one as a base for a Dashboard Variable. That would allow users to pick from the list of properties for an asset within a panel. 
We could use a Line chart to visualize all properties (time series) from an asset instead of creating bespoke panels for specific properties ahead of time. 

**Additional context**
We're using ```v1.16.1```.

### LINKED ISSUE #412: Support multiple properties in data queries
For the IoT SiteWise Get* queries (GetPropertyValue, GetPropertyValueHistory, GetInterpolatedValues, and GetPropertyAggregates) we support selecting multiple assets in the same asset model, but we do not support selecting multiple properties for those assets. Many IoT SiteWise customers are asking for this feature.

Like we batch calls to the Get* APIs for multiple assets, we should have additional entries in the batch request for multiple properties.

Each entry to a batch data call must have at least an assetId and property OR a property alias. The scope for this feature is to support a list of assetIds and propertyIds and return data for all combinations of those ids. Every combination is valid as long as each asset is an instance of the same asset model.

The backend plugin will need to update models to accept an array of propertyIds like assetIds, then create entries for the respective batch API like here: https://github.com/grafana/iot-sitewise-datasource/blob/main/pkg/sitewise/api/property_history_batch.go#L48

The frontend plugin will need to update the propertyId field to support multiple values.

The only alternative is to create a new query for each property, which is not extensible for customers with many properties on their assets.

**Additional context**
N/A

hiralpatelHM:
…ldcard queries

Added ability to select multiple properties from an asset along with using  wildcards as well, making data queries easier and more flexible in the Visual Query Builder.

Fixes #412  
Fixes #283  

Implemented wildcard handling in property selection.
If a user searches with a wildcard (e.g., `*power*`), all matching properties will be automatically selected.

### DIFF
--- a/src/components/query/visual-query-builder/PropertyQueryEditor.tsx
+++ b/src/components/query/visual-query-builder/PropertyQueryEditor.tsx
@@ -153,7 +153,9 @@ export const PropertyQueryEditor = ({ query, datasource, onChange }: SitewiseQue
         if (propertyId.includes('*') && assetProperties.length) {
           const regex = new RegExp(`^${propertyId.replace(/[.+?^${}()|[\]\\]/g, '\\$&').replace(/\*/g, '.*')}$`, 'i');
           assetProperties.forEach(({ label, value }) => {
-            if (label && regex.test(label) && value) existingIds.add(value);
+            if (label && regex.test(label) && value) {
+              existingIds.add(value);
+            }
           });
           return onChange({ ...query, propertyIds: [...existingIds] });
         }


chejimmy:
> Hi @hiralpatelHM thanks for the contribution!
> 
> The wildcard implementation looks good.
> 
> I might be misunderstanding, but I don't think this PR solves #412, or at least not in the way I would expect it to. It's true that now we can select multiple properties, but if I try to select multiple assets and their respective properties, the asset property dropdown acts strange - it does't show me all properties from ALL assets selected, but only from the first asset selected. Are you getting the same thing? The description of [this PR](https://github.com/grafana/iot-sitewise-datasource/pull/435) says
> 
> > Update (25/03/06): single property id selection in UI until a better UX is designed around pairing assets and properties
> 
> The issue is if we enable selection of multiple properties for wildcard queries, we are accepting the bug described above. @chejimmy do you have more info on this - is this ok to merge with this limitation?

TY for pointing out the issue; we've moved this discussion to Slack 🙏 Pending decision at the moment :)

idastambuk:
@hiralpatelHM looks good now, just yarn lint failing:
Screenshot 2025-09-01 at 12 21 25

PR #644, opened 2025-11-28
Implement JSON string parsing to extract anomaly-related fields for property query type

hiralpatelHM:
- Implement JSON parsing in ParseJSONFields to extract anomaly-related fields
- Parse top-level anomaly_score, prediction, and prediction_reason
- Parse diagnostics array to extract contribution values and normalize percentages

Fixes #Anomaly Detection

### DIFF
--- a/pkg/sitewise/json_mapping.go
+++ b/pkg/sitewise/json_mapping.go
@@ -37,7 +37,6 @@ var requiredJSONFields = []string{
 //   - assetID is empty
 //   - the asset model cannot be fetched
 //   - no properties are found in the asset model
-
 func BuildPropertyNameMap(
 	ctx context.Context,
 	resources resource.ResourceProvider,
@@ -54,7 +53,6 @@ func BuildPropertyNameMap(
 	}
 
 	propertyNameMap := make(map[string]string)
-
 	for _, prop := range modelResp.AssetModelProperties {
 		if prop.Id != nil && prop.Name != nil {
 			propertyNameMap[*prop.Id] = *prop.Name
@@ -72,7 +70,6 @@ func BuildPropertyNameMap(
 }
 
 // requiresJsonParsing determines whether a given query type
-
 func requiresJsonParsing(query models.BaseQuery) bool {
 	switch query.QueryType {
 	case models.QueryTypePropertyValueHistory,
@@ -98,7 +95,7 @@ func ParseJSONFields(
 	resources resource.ResourceProvider,
 	assetID string,
 ) data.Frames {
-	backend.Logger.Info("ParseJSONFields: starting JSON parsing", "assetID", assetID)
+	backend.Logger.Debug("ParseJSONFields: starting JSON parsing", "assetID", assetID)
 
 	// Build property ID -> readable name map once per request
 	propertyNameMap, err := BuildPropertyNameMap(ctx, resources, assetID)
@@ -134,10 +131,6 @@ func ParseJSONFields(
 			for r := 0; r < rowCount; r++ {
 				rawStr, ok := field.At(r).(string)
 				if !ok || rawStr == "" {
-					continue
-				}
-				// Perform a fast check to ensure the value looks like a JSON object
-				if !strings.HasPrefix(rawStr, "{") {
 					continue
 				}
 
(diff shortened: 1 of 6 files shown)


pkg/sitewise/json_mapping.go
```diff
@@ -0,0 +1,182 @@
...
+				}
+
+				var obj map[string]interface{}
+				if err := json.Unmarshal([]byte(rawStr), &obj); err != nil {
+					backend.Logger.Warn("ParseJSONFields: corrupted JSON, skipping row", "err", err, "frame", frame.Name, "row", r)
+					continue
+				}
+				jsonParsed = true
+				// Required fields check
+				for _, req := range []string{"timestamp", "prediction", "prediction_reason"} {
```
iwysiu: The required fields should be consts since we use them multiple times

pkg/sitewise/json_mapping.go
```diff
@@ -0,0 +1,182 @@
...
+					if _, ok := obj[req]; !ok {
+						backend.Logger.Warn("ParseJSONFields: missing required field", "field", req, "frame", frame.Name, "row", r)
+					}
+				}
+
+				for key, val := range obj {
+					if key == "diagnostics" || key == "timestamp" {
```
iwysiu: shouldn't we also skip anomaly score if we handle it in the next block? Also are we intentionally skipping timestamp?

pkg/sitewise/json_mapping.go
```diff
@@ -0,0 +1,182 @@
...
+	"github.com/grafana/grafana-plugin-sdk-go/data"
+	"github.com/grafana/iot-sitewise-datasource/pkg/models"
+	"github.com/grafana/iot-sitewise-datasource/pkg/sitewise/resource"
+)
+
+func BuildPropertyNameMap(ctx context.Context, resources resource.ResourceProvider, assetID string) map[string]string {
+	propertyNameMap := map[string]string{}
+	if assetID == "" {
+		backend.Logger.Warn("BuildPropertyNameMap: empty assetId")
+		return propertyNameMap
```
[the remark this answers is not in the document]
hiralpatelHM: @iwysiu I’ve updated the flow so that BuildPropertyNameMap returns an error when the assetID is empty instead of returning an empty map. In ParseJSONFields, we now handle this explicitly by logging the error and exiting early

pkg/sitewise/json_mapping.go
```diff
@@ -0,0 +1,254 @@
...
+						parts := strings.Split(rawName, "\\")
+						if len(parts) < 2 {
+							continue
+						}
+						propertyID := parts[1]
+						readable := propertyID
+						if mapped, ok := propertyNameMap[propertyID]; ok {
+							readable = mapped
+						}
+						fieldName := "contrib_" + assetName + "_" + readable
```
ahammedf: The label is not being generated correctly. The readable variable contains the ID rather than the property name. See the series labels in the screenshot

Screenshot 2026-01-09 at 3 19 39 PM

ahammedf: Per our offline discussion, we can leave the code as-is and handle this use case in a new commit. This PR can focus on parsing the json input

pkg/sitewise/json_mapping.go
```diff
@@ -0,0 +1,254 @@
...
+			"BuildPropertyNameMap: no properties found for assetId %s",
+			assetID,
+		)
+	}
+
+	return propertyNameMap, nil
+}
+
+// requiresJsonParsing determines whether a given query type
+
```
iwysiu: nit: remove empty line

pkg/sitewise/json_mapping.go
```diff
@@ -0,0 +1,254 @@
...
+			rowCount := field.Len()
+			jsonFields := map[string]*data.Field{}
+
+			for r := 0; r < rowCount; r++ {
+				rawStr, ok := field.At(r).(string)
+				if !ok || rawStr == "" {
+					continue
+				}
+				// Perform a fast check to ensure the value looks like a JSON object
+				if !strings.HasPrefix(rawStr, "{") {
```
iwysiu: This could probably be removed? since it would error for `json.Unmarshal` anyway and I'm assuming that would handle this case

iwysiu (approved):
A couple style nits, but overall looks good!

I rebased because it was running into an issue in one of the checks that I think rebasing should fix.

iwysiu:
Merged in https://github.com/grafana/iot-sitewise-datasource/pull/704

PR #723, opened 2026-02-24
fix(json): use caching to retrieve asset and property names correctly

ahammedf:
Export CachingResourceProvider to make it accessible from json_mapping and use it to fetch asset and property names when parsing anomaly detection JSON data, ensuring names are retrieved correctly from multiple assets if required.

This fixes the issue where diagnostic field names were not being resolved correctly when processing anomaly detection data from multiple assets.

Fixes the issue mentioned here: https://github.com/grafana/iot-sitewise-datasource/pull/644#discussion_r2677917333


kevinwcyu (changes requested):
Hi @ahammedf,

Thanks for working on this. I think the overall fix makes sense, but I’d suggest avoiding exporting the caching resource provider just so `ParseJSONFields` can resolve diagnostic asset/property names.

I pushed a branch showing what I mean here:

https://github.com/grafana/iot-sitewise-datasource/compare/anomaly_detection_multiple_assets...kevinwcyu/pr-723-comments

The main change is to keep the caching provider internal and expose a small lookup interface from the query resource provider instead. That lets the JSON parser resolve arbitrary SiteWise resources from the diagnostic payload without depending directly on the concrete cache implementation.

The shape is roughly:

```go
type ResourceLookup interface {
	LookupAsset(ctx context.Context, assetId string) (*iotsitewise.DescribeAssetOutput, error)
	LookupAssetProperty(ctx context.Context, assetId string, propertyId string, propertyAlias string) (*iotsitewise.DescribeAssetPropertyOutput, error)
	LookupAssetModel(ctx context.Context, modelId string) (*iotsitewise.DescribeAssetModelOutput, error)
}
```

Then `ParseJSONFields` can take `resource.ResourceLookup`, and `response_processor.go` can pass the existing query resource provider instead of the concrete caching provider.

This keeps the cache as an implementation detail while still supporting diagnostic labels that refer to asset/property pairs outside the original query scope.

pkg/sitewise/json_mapping.go
```diff
@@ -88,30 +49,13 @@ func requiresJsonParsing(query models.BaseQuery) bool {
...
-//   - Diagnostic contribution values are normalized to percentages
 func ParseJSONFields(
 	ctx context.Context,
 	frames data.Frames,
-	resources resource.ResourceProvider,
-	assetID string,
+	resources resource.ResourceLookup,
 ) data.Frames {
-	backend.Logger.Debug("ParseJSONFields: starting JSON parsing", "assetID", assetID)
+	backend.Logger.Debug("ParseJSONFields: starting JSON parsing", "frames", frames)
```
kevinwcyu:
```suggestion
```

let's remove this logline, the frames can be quite large and can contain customer data that shouldn't be logged.

kevinwcyu (commented):
lgtm, just one more minor change.

kevinwcyu (approved):

pkg/sitewise/json_mapping.go
ahammedf: Removed debug log

PR #756, opened 2026-03-27
chore: add debug logs for raw property alias fallback

kevinwcyu:
## Summary
- Add debug log lines when SiteWise alias resolution falls back to raw property alias metadata
- Add a debug log line when cached fallback metadata is reused
- These additional debug log lines are for https://github.com/grafana/support-escalations/issues/21107

## Files Changed
- `pkg/resource/sitewise.go`
- `pkg/resource/caching_resource_provider.go`


adamyeats (approved):
`isRawPropertyAliasFallback` could use a unit test but otherwise LGTM.

PR #808, opened 2026-06-25
Fix 400 error when batched IoT SiteWise queries paginate past completed entries

### LINKED ISSUE #765: Empty nextToken sent during BatchGetAssetPropertyValueHistory pagination causes 500 error
**What happened**:

   When querying IoT SiteWise asset property value history with a time range exceeding approximately 3 hours, the plugin returns a 500 error. The underlying cause is the plugin sending an empty string `nextToken: ""`
   during pagination of the `BatchGetAssetPropertyValueHistory` API. The IoT SiteWise API rejects this with a 400 `InvalidRequestException`:

   Status: 500. Message: failed to fetch query data: operation error IoTSiteWise:
   BatchGetAssetPropertyValueHistory, https response error StatusCode: 400,
   RequestID: cb509df8-8cfd-42c4-b80a-76cf07a306bd,
   InvalidRequestException: 2 validation errors detected:
     Value '' at 'nextToken' failed to satisfy constraint: Member must satisfy
     regular expression pattern: [A-Za-z0-9+/=]+;
     Value '' at 'nextToken' failed to satisfy constraint: Member must have length
     greater than or equal to 1

   Queries with time ranges exceeding 3 hours should paginate correctly. The plugin should pass the valid `nextToken` from the previous API response, or omit the parameter entirely on the first request — not send an
   empty string.

   1. Configure an IoT SiteWise datasource
   2. Create a panel querying asset property value history
   3. Set the time range to "Last 6 hours" (any range > ~3 hours that requires pagination)
   4. Observe the 500 error
   5. Change time range to "Last 3 hours" or less — query succeeds (no pagination needed)

   **Screenshots**

   HAR file analysis confirms the plugin sends `nextToken: ""` in the paginated request payload. HAR file available on request.

   - The issue only occurs when the result set is large enough to require pagination (typically > 3 hour time ranges)
   - Queries with time ranges ≤ 3 hours work correctly because all data fits in a single API response page
   - IoT SiteWise API `nextToken` validation requires: pattern `[A-Za-z0-9+/=]+`, minimum length 1 — empty strings are rejected
   - API docs: https://docs.aws.amazon.com/iot-sitewise/latest/APIReference/API_BatchGetAssetPropertyValueHistory.html
   - Workaround: use time ranges of 3 hours or less

   **Environment**:

   - Grafana version: Any version
   - Plugin version: 2.4.2
   - OS Grafana is installed on: Any OS
   - User OS & Browser: N/A (server-side plugin error)
   - Others:

ashishranjan738:

`getNextToken` returned `aws.String("")` for batched time-series queries when an entry's token was absent from the `NextTokens` map (a map miss). This happens once an entry has completed paginating on an earlier page: the frontend (`src/getNextQueries.ts`) only writes `nextTokens[entryId]` when the token is truthy, so the completed entry drops out of the map, and on the next continuation call the backend recomputes that `entryId`, misses, and gets the Go zero value `""`.

A non-nil pointer to an empty string is serialized onto `BatchGetAssetPropertyValueHistoryInput.NextToken` and forwarded verbatim by the SDK paginator on its first request, so IoT SiteWise rejects it:

```
InvalidRequestException: 2 validation errors detected:
Value '' at 'nextToken' failed to satisfy constraint:
Member must satisfy regular expression pattern: [A-Za-z0-9+/=]+;
Value '' at 'nextToken' failed to satisfy constraint:
Member must have length greater than or equal to 1
```

This surfaces as an HTTP 500 in Grafana for any panel whose time range is large enough to require multi-page pagination (commonly > ~3 hours). It affects all batched query paths that share `getNextToken` (property history, property value, property aggregate).

The fix mirrors the existing single-entry branch, which already guards `query.NextToken == ""` → `return nil`. Returning `nil` omits the token, which correctly ends pagination for that entry instead of sending an invalid request. `getNextToken` is byte-identical on `main` and tags v2.4.2 / v2.5.1 / v2.6.0, so this bug reproduces on every currently published version.

Fixes #765

- Adds `pkg/sitewise/api/util_internal_test.go` (internal test, since `getNextToken` is unexported) covering both branches: single-entry (empty/non-empty) and batched (non-empty, map-miss, explicit-empty, alias-keyed). The map-miss and explicit-empty cases fail without this change.
- No behavior change for the non-empty token case — only an empty resolved token now ends pagination (returns `nil`) instead of producing a guaranteed-invalid request.
- Adding the `changelog` label as this is a user-facing bug fix.

### DIFF
--- a/pkg/sitewise/api/util.go
+++ b/pkg/sitewise/api/util.go
@@ -35,6 +35,13 @@ func getNextToken(query models.BaseQuery) *string {
 		}
 		// If there are any issues looking up the nextToken it should error and bubble up
 		nextToken := query.NextTokens[entryId]
+		// A map miss (e.g. an entry that completed pagination on a prior page) yields
+		// the empty string. Sending an empty nextToken to the IoT SiteWise API is
+		// rejected with a 400 InvalidRequestException, so omit it (return nil) to end
+		// pagination for that entry, mirroring the single-entry branch below.
+		if nextToken == "" {
+			return nil
+		}
 		return aws.String(nextToken)
 	} else {
 		if query.NextToken == "" {
--- a/pkg/sitewise/api/util_internal_test.go
+++ b/pkg/sitewise/api/util_internal_test.go
@@ -1,0 +1,112 @@
+package api
+
+import (
+	"testing"
+
+	"github.com/grafana/iot-sitewise-datasource/pkg/models"
+	"github.com/grafana/iot-sitewise-datasource/pkg/util"
+)
+
+func TestGetNextToken(t *testing.T) {
+	const (
+		assetID    = "11111111-1111-1111-1111-111111111111"
+		propertyID = "22222222-2222-2222-2222-222222222222"
+		alias      = "/my/property/alias"
+	)
+	batchedEntryID := *util.GetEntryIdFromAssetProperty(assetID, propertyID)
+	aliasEntryID := *util.GetEntryIdFromPropertyAlias(alias)
+
+	tests := []struct {
+		name  string
+		query models.BaseQuery
+		want  *string
+	}{
+		{
+			name:  "no token returns nil",
+			query: models.BaseQuery{},
+			want:  nil,
+		},
+		{
+			name:  "single-entry non-empty token is forwarded",
+			query: models.BaseQuery{NextToken: "abc123"},
+			want:  stringPtr("abc123"),
+		},
+		{
+			name:  "single-entry empty token returns nil",
+			query: models.BaseQuery{NextToken: ""},
+			want:  nil,
+		},
+		{
+			name: "batched entry with non-empty token is forwarded",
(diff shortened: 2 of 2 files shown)


akkikumar72:
👋 Hi @ashishranjan738,  thank you for the contribution.

Thanks for working on this. I think the overall fix makes sense, few changes before we merge it

- Please sign the CLA 
-  There is some policy changes, could you recommit your changes in signed commits?
🛈 Here is some info on [signed commits within github](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits).
- We can run the workflow checks again

akkikumar72 (commented):
Few comments, before we can merge it

akkikumar72:
Closing this PR and recreating new one - https://github.com/grafana/iot-sitewise-datasource/pull/831, so we can merge it sooner

PR #827, opened 2026-07-23
Resolve critical CVEs

mattvella07:

Fixes #


aangelisc (commented):
Same as Athena - let's not add resolutions 😊

PR #831, opened 2026-07-28
fix(api): don't send empty nextToken when batched IoT SiteWise queries finish paginating

### LINKED ISSUE #765: Empty nextToken sent during BatchGetAssetPropertyValueHistory pagination causes 500 error

akkikumar72:
- Recreates #808 (thanks @ashishranjan738 for the original fix) with signed commits so it can pass CI/CLA — the original author is no longer active to address review feedback.
- `getNextToken` returned `aws.String("")` on a `NextTokens` map miss (an entry that finished paginating on a prior page); IoT SiteWise rejects an empty `nextToken` with a 400, surfacing as a 500 in Grafana for any time range needing pagination (~3h+). Now returns `nil` to omit the token, mirroring the existing single-entry guard.
- Also clarifies a stale comment flagged in Copilot's review of #808.
- Checks: `go build ./...`, `go vet ./...`, `go test ./pkg/...` all pass, including new regression tests for #765.

Fixes #765

### DIFF
--- a/pkg/sitewise/api/util.go
+++ b/pkg/sitewise/api/util.go
@@ -33,8 +33,16 @@ func getNextToken(query models.BaseQuery) *string {
 		} else {
 			entryId = *util.GetEntryIdFromPropertyAlias(entry.PropertyAlias)
 		}
-		// If there are any issues looking up the nextToken it should error and bubble up
+		// A map miss (e.g. an entry that completed pagination on a prior page) yields
+		// the empty string, same as an explicitly empty token. Sending an empty
+		// nextToken to the IoT SiteWise API is rejected with a 400
+		// InvalidRequestException, so treat both cases as end-of-pagination for
+		// that entry and omit the token (return nil), mirroring the single-entry
+		// branch below.
 		nextToken := query.NextTokens[entryId]
+		if nextToken == "" {
+			return nil
+		}
 		return aws.String(nextToken)
 	} else {
 		if query.NextToken == "" {
--- a/pkg/sitewise/api/util_internal_test.go
+++ b/pkg/sitewise/api/util_internal_test.go
@@ -1,0 +1,112 @@
+package api
+
+import (
+	"testing"
+
+	"github.com/grafana/iot-sitewise-datasource/pkg/models"
+	"github.com/grafana/iot-sitewise-datasource/pkg/util"
+)
+
+func TestGetNextToken(t *testing.T) {
+	const (
+		assetID    = "11111111-1111-1111-1111-111111111111"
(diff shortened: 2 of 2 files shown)


njvrzm (approved):

akkikumar72:
Hey @ashishranjan738, we are missing  CLA signature. Can you please click above comment from `cla-assistant` and approve it. So, I can take it forward.

PR #837, opened 2026-08-06
fix(security): bump out-of-SLO serialize-javascript to 7.0.7

akkikumar72:
- Pin out-of-SLO `serialize-javascript` 6.0.2 -> 7.0.7 via selective resolution (GHSA-5c6j-r48x-rmvq, CVE-2026-34043)
- Dedupes onto the 7.x entry already in yarn.lock; leaves in-SLO deps alone
- Checks: `yarn typecheck`, `yarn test:ci` (25 suites, 204 tests passed)

### DIFF
--- a/package.json
+++ b/package.json
@@ -100,7 +100,8 @@
     "webpack-virtual-modules": "^0.6.2"
   },
   "resolutions": {
-    "diff": "8.0.3"
+    "diff": "8.0.3",
+    "serialize-javascript@npm:^6.0.2": "7.0.7"
   },
   "packageManager": "yarn@4.12.0"
 }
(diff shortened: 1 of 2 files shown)


ryanmorris708 (approved):

ryanmorris708 (changes requested):
Can we bump the version for `terser-webpack-plugin` instead? `5.6.1` has removed the dependency on `serialize-javescript` entirely

akkikumar72:
> Can we bump the version for `terser-webpack-plugin` instead? `5.6.1` has removed the dependency on `serialize-javescript` entirely

Good call, done. Bumped terser-webpack-plugin to 5.6.1 and dropped the resolution entirely.

ryanmorris708 (approved):

